PR #8, opened 2017-10-08
Use extras for optional dependencies

### LINKED ISSUE #7: Use extras for optional dependencies
Currently, selenium, pymysql, and psycopg2 are all installed regardless of what I need.

Extras should be used like [sqlalchemy does](https://github.com/zzzeek/sqlalchemy/blob/2c280791886756422a8103769cf131b0fe292ffe/setup.py#L156-L165).

That way, I can have `testcontainers[postgresql,mysql]` as a dependency if that's all I need.

I can submit a PR if you like.

RazerM:
Closes #7

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,13 +23,17 @@ Currently available features:
 Installation
 ------------
 
-The **testcontainers** module is available from PyPi at:
+The **testcontainers** module is available from PyPI at:
 
 * https://pypi.python.org/pypi/testcontainers
 
-and can be installed using ``pip``.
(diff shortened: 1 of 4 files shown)


SergeyPirogov:
There is a broken build https://travis-ci.org/SergeyPirogov/testcontainers-python/builds/285263356?utm_source=github_status&utm_medium=notification, but from the error message, I can see that the root cause probably bad Docker installation. Need to investigate. Will merge this when the build and all the tests will pass

SergeyPirogov:
@RazerM could you please fix the conflict

PR #12, opened 2018-01-07
volumes removal and fixing CI

t0ch1k:
- now by default containers will be deleted with their volumes after test session
- version 2.1.2
- removed codecov from travis script. Anyway it never worked

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,8 @@ install:
 
 script:
   - pipenv install --dev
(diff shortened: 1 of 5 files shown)


SergeyPirogov:
@t0ch1k please fix conflict

t0ch1k:
I don't understand why it fails now. I've removed coverage from build script, but still it fails in case of coverage

SergeyPirogov:
Will look at it, maybe codecov is broken

PR #19, opened 2018-05-04
Add elasticsearch container

nivm:
Add an implementation of the elasticsearch container.


SergeyPirogov:
@nivm Thank you for PR, could you please add some tests for elastic search container? See example for DB or Selenium

PR #23, opened 2018-08-11
Add support for docker-outside-of-docker

tcwalther:
This commit adds support for running `testcontainers` inside Docker using the so-called "docker-outside-of-docker" approach: by mounting `/var/run/docker.sock` into the Docker container (using, for example, `docker run -v /var/run/docker.sock:/var/run/docker.sock ...`), one can spawn the test containers as sibling-containers.

To determine whether `testcontainers` is run inside a container, it checks whether a `/.dockerenv` file exists.

This PR also adds a simple `Dockerfile` that one can use to for testing. I have not, however, updated the `.travis.yml` to test this in CI - I'm open for suggestions on how to do so.

It also includes a small fix I made to make sure that Docker containers are stopped even when an exception is thrown:

```python
    # in class DockerContainer:
    def __del__(self):
        """
        Try to remove the container in all circumstances
        """
        if self._container is not None:
            try:
                self.stop()
            except:
                pass
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM python:3.6
+
+COPY ./Pipfile* /workspace/
+WORKDIR /workspace
+
+RUN pip install pipenv && \
+    pipenv install --system --dev --deploy
+
+COPY . /workspace
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,20 +1,7 @@
 {
     "_meta": {
         "hash": {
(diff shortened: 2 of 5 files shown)


SergeyPirogov:
Looks good, thank you for this PR

tcwalther:
@SergeyPirogov thanks a lot for merging. Would you be happy to publish a new version 2.3 on PyPI that includes this change?

SergeyPirogov:
It has been published a few days ago

tcwalther:
That’s amazing, thanks a lot!

PR #34, opened 2019-08-27
Allow passing in any parameters supported by docker

weixu365:
Allow passing in any parameters supported by docker, so that user can pass in 
- New parameters supported by docker 
- Existing parameters but not supported by testcontainers, e.g. there are almost hundred parameters at: https://docker-py.readthedocs.io/en/stable/containers.html

In my own scenario, I need to start a container which shares the same volume from another container (not created by testcontainer).

### DIFF
--- a/testcontainers/core/container.py
+++ b/testcontainers/core/container.py
@@ -8,7 +8,7 @@ from testcontainers.core.utils import is_windows, inside_container
 
 
 class DockerContainer(object):
-    def __init__(self, image):
+    def __init__(self, image, **kargs):
         self.env = {}
         self.ports = {}
         self.volumes = {}
@@ -17,6 +17,7 @@ class DockerContainer(object):
         self._container = None
         self._command = None
         self._name = None
+        self._kargs = kargs
 
     def with_env(self, key: str, value: str) -> 'DockerContainer':
         self.env[key] = value
@@ -32,6 +33,10 @@ class DockerContainer(object):
             self.ports[port] = None
         return self
 
+    def with_kargs(self, **kargs) -> 'DockerContainer':
+        self._kargs = kargs
+        return self
+
     def start(self):
         print("")
         print("{} {}".format(crayons.yellow("Pulling image"),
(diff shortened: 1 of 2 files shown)


testcontainers/core/container.py
```diff
@@ -8,7 +8,7 @@
 
 
 class DockerContainer(object):
-    def __init__(self, image):
+    def __init__(self, image, **kargs):
```
tillahoffmann: The standard is to use `kwargs` rather than `kargs`.

testcontainers/core/container.py
```diff
@@ -44,7 +49,9 @@ def start(self):
                                                 environment=self.env,
                                                 ports=self.ports,
                                                 name=self._name,
-                                                volumes=self.volumes)
+                                                volumes=self.volumes,
+                                                **self._kargs
```
tillahoffmann: This may raise a `TypeError` if one of the arguments specified explicitly also appears in the keyword arguments. For example,  the following will fail.

```python
DockerContainer('hello-world', image='ubuntu').start()
```
```

File: testcontainers/core/container.py [hunk 3]
```diff
@@ -32,6 +33,10 @@ def with_exposed_ports(self, *ports) -> 'DockerContainer':
             self.ports[port] = None
         return self
 
+    def with_kargs(self, **kargs) -> 'DockerContainer':
+        self._kargs = kargs
```
tillahoffmann: Is replacing the entire dictionary the desired behaviour? E.g. the current implementation does not allow the keyword arguments to be updated.

SergeyPirogov:
@tillahoffmann So as I understand you propose to delete methods such as ports and volumes and replace them with kwargs?

weixu365:
I think it's better to keep the methods and only remove the fields, so keep everything backward compatible, I'm creating another PR now

testcontainers/core/container.py
weixu365: I think this can't be solved unless removing the `image` parameter which breaks the existing usages.

PR #40, opened 2019-09-27
Use binary psycopg2 packages

walles:
Quoting [the `psycopg2-binary` web page](https://pypi.org/project/psycopg2-binary/):
> The binary package is a practical choice for development and testing but in production it is advised to use the package built from sources.

For me right now, not having had to build `psycopg2` from source while testing would have made my life easier.


walles:
@SergeyPirogov would it be possible for you to make a new release with this change in it?

PR #49, opened 2020-01-03
Solved import error in mysql example

dabrign:
Solvin an import error, increment Mysql version to match last no-deprecated image on dockerhub

### DIFF
--- a/tests/examples/mysql_example.py
+++ b/tests/examples/mysql_example.py
@@ -1,12 +1,12 @@
 import pytest
 import sqlalchemy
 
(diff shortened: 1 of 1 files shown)


tillahoffmann:
On second thought, it looks like the example is redundant given the tests [here](https://github.com/testcontainers/testcontainers-python/blob/master/tests/test_db_containers.py). 

@SergeyPirogov, thoughts on deleting the example to avoid confusion/duplication?

GregHilston:
If it means anything, this pull request was really helpful for quickly getting up and running

PR #55, opened 2020-01-16
Add option to wait for logs.

tillahoffmann:
Reduce test time by waiting for specific log message rather than suspending execution. Cf #52.

PR #56, opened 2020-01-23
added Neo4j container

eastlondoner:
Let me know if you'd like anything else for this.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -16,6 +16,7 @@ Currently available features:
 * Selenium Standalone containers
 * MySql Db container
 * MariaDb container
+* Neo4j container
 * OracleDb container
 * PostgreSQL Db container
 * Microsoft SQL Server container
--- a/docs/database.rst
+++ b/docs/database.rst
@@ -1,7 +1,7 @@
 Database containers
 ===================
 
-Allows to spin up database images such as MySQL, PostgreSQL, MariaDB, Oracle XE, or MongoDb.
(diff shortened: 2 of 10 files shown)


eastlondoner:
I've implemented my own equivalent of  `wait_for_logs` but can switch to use the one from https://github.com/testcontainers/testcontainers-python/pull/55 when available

tillahoffmann:
@eastlondoner, #55 has now been merged if you want to use that version of `wait_for_logs`.

eastlondoner:
@tillahoffmann thanks, I've updated the PR :-)

eastlondoner:
@tillahoffmann I've rebased, updated the code and added basic docs.

I re-built the requirements files which pulled in a load of updated dependencies, I hope that's OK.

eastlondoner:
@tillahoffmann so that requirements update makes the docker client too new for the CI server. So I'll pin `docker-compose` for now ... new commit incoming.

tillahoffmann:
Yes, it looks a bit fiddly. It looks like it might be `docker` rather than `docker-compose`. Fancy adding the pinned version to `requirements.in` rather than `setup.py` (that'll mean people can still install a newer version on their machine if they want)?

eastlondoner:
OK. I manually pinned the docker version in `requirements.in` Looks like CI is working now :-)

tillahoffmann:
Great! I'll try to send you a PR to the PR to make sure we don't have the same problem in the future. Shouldn't be long.

tillahoffmann:
https://github.com/eastlondoner/testcontainers-python/pull/1

Issue #38, opened 2019-09-13
get_container_host_ip returning 0.0.0.0 instead of localhost


rnorth:
When run locally, we're having trouble using `get_container_host_ip`, as it is returning `0.0.0.0`.

https://github.com/testcontainers/testcontainers-python/blob/dc956ccd2f7ee80964201eba2f713d6b3379b74c/testcontainers/core/container.py#L87

As the container IP is intended to be used by clients, surely it should be returning either `localhost` or `127.0.0.1`? I'm not sure about other Python clients, but with asyncio ClientSession we're finding that it is unwilling to accept `0.0.0.0` as a valid server address.

I'd be happy to raise a PR, but again don't want to cause breakage if there's a specific reason for it to be as it is.

cal97g:
@rnorth `0.0.0.0` means it's listening on any possible interface on the local machine iirc - see this: https://serverfault.com/a/78058

If asyncio doesn't like 0.0.0.0, you can use `127.0.0.1` and you'll get the desired effect.

rnorth:
@cal97g that's what I thought; I'm quite used to seeing `0.0.0.0` as a server address but never from the client side.

I think on the basis that `127.0.0.1` is _at least_ as valid an address from a client's perspective, I'd like it if we could contribute a change.

cal97g:
@rnorth I'd suggest just opening a pull request for this and your other issue with a fix. The maintainer doesn't seem to have much free time but tends to be more reactive to PR's in my experience.

PR #59, opened 2020-02-05
Use correct hostname in ElasticsearchContainer

ahaeger:
My tests are running from within a docker container and noticed this issue, as it could never connect to Elasticsearch.


ahaeger:
Travis seem to have some connection problems.

tillahoffmann:
Thanks for this. Let's merge #61 first so we don't break something (see #38 for details of the bug).

Issue #20, opened 2018-05-21
[Documentation] Python version compatability should be indicated.


bearrito:
I believe this library only works with Python 3+, this isn't indicated in the docs though.

tillahoffmann:
Resolved in #66.

PR #63, opened 2020-02-15
Require python >= 3.4

martykube:
This will enforce minimum python version during installation.  

You have to have up-to-date versions of pip and setuptools for this to be enforced: https://python3statement.org/practicalities/

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -4,6 +4,8 @@ verify_ssl = true
 
 ["dev-packages"]
 codecov = "*"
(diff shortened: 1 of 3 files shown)


martykube:
Resolves #20

martykube:
To enforce python3 we need to use recent build tools when preparing the distribution. 

 This change makes sure we have pip>=9.0 and setuptools >=24.2.0 for 'pipenv install --dev'.

Issue #68, opened 2020-02-25
ServerSelectionTimeoutError when running in gitlab-ci on gitlab.com


vinjana:
I am running testcontainers-python 2.6 with mongo:4.2.3. My tests work fine on my local system, but when I have them run on gitlab.com all testcontainer-using tests fail with errors of the following type:

```
557 ___________________ ERROR at teardown of test_create_new_run ___________________
558     @pytest.fixture(scope="function")
559     def database_connection():
560         container = MongoDbContainer('mongo:4.2.3')
561         container.start()
562         print("MongoDB connection URL: " + container.get_connection_url())
563         database = Database(MongoClient(container.get_connection_url()), "WES_Test")
564         yield database
565 >       database._db_runs().drop()
566 tests/Database_test.py:14: 
567 _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
568 /opt/conda/envs/wesnake/lib/python3.7/site-packages/pymongo/collection.py:1103: in drop
569     dbo.drop_collection(self.__name, session=session)
570 /opt/conda/envs/wesnake/lib/python3.7/site-packages/pymongo/database.py:914: in drop_collection
571     with self.__client._socket_for_writes(session) as sock_info:
572 /opt/conda/envs/wesnake/lib/python3.7/site-packages/pymongo/mongo_client.py:1266: in _socket_for_writes
573     server = self._select_server(writable_server_selector, session)
574 /opt/conda/envs/wesnake/lib/python3.7/site-packages/pymongo/mongo_client.py:1253: in _select_server
575     server = topology.select_server(server_selector)
576 /opt/conda/envs/wesnake/lib/python3.7/site-packages/pymongo/topology.py:235: in select_server
577     address))
578 /opt/conda/envs/wesnake/lib/python3.7/site-packages/pymongo/topology.py:193: in select_servers
579     selector, server_timeout, address)
580 _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
581 self = <pymongo.topology.Topology object at 0x7fc6ab1a8c88>
582 selector = <function writable_server_selector at 0x7fc6ae81af28>, timeout = 30
583 address = None
584     def _select_servers_loop(self, selector, timeout, address):
[...]

Issue #60, opened 2020-02-11
v2.5 installed from PyPI does not have mongodb module


dsabsay:
I've installed v2.5 via `pip`. Attempting to use the `mongodb` module like so:
```
from testcontainers.mongodb import MongoDbContainer
```
gives me:
```
ImportError: No module named mongodb
```
Looking in my virtual environment:
```
(env) sabsays-MacBook-Pro:st2-test-utils sabsay$ ls env/lib/python2.7/site-packages/testcontainers/
__init__.py       compose.pyc       elasticsearch.pyc google            nginx.py          oracle.pyc        redis.py
__init__.pyc      core              general.py        mysql.py          nginx.pyc         postgres.py       redis.pyc
compose.py        elasticsearch.py  general.pyc       mysql.pyc         oracle.py         postgres.pyc      selenium.py
```
it is indeed not there :-)

dsabsay:
Oops, hehe. Forgot about the options to install with `pip install testcontainers[mongodb]`.

That seems to work although, `core/generic.py` imports sqlalchemy which shouldn't be needed to use mongodb. Think the `DbContainer` class in core/generic.py really doesn't belong there, as it brings in extra (optional) dependencies.

vinjana:
I still have to use

```bash
pip install git+https://github.com/testcontainers/testcontainers-python.git` 
```

to get a working testcontainers[mongodb] in Gitlab-CI.

tillahoffmann:
The extra is called `mongo` rather than `mongodb`. Version 2.6.0 should have that included.

Issue #43, opened 2019-11-28
testcontainers wait indefinitely without effects when running in Gitlab-CI


victornoel:
I'm running a test based on testcontainers-python in a container running in Gitlab CI.

Basically the CI spawns a docker-in-docker container next to the container running the tests. I then set the environment variable DOCKER_HOST so that it can be used by testcontainers.

But in practice the test just get stuck, so I ran it with `pytest -s` to see the logs of testcontainers and got:
```
tests/test.py
Pulling image postgres:10
Container started:  4a1115dc98
Waiting to be ready...
```

And then it gets stuck indefinitely (I waited 20 minutes).

For the record, testcontainers-java works perfectly with this same situation, so it should be possible to make this work.

I'm not sure how I can get more debug information to understand what is happening, but shouldn't it timeout normally at least?

victornoel:
See also #25 for a different but maybe related problem and https://www.testcontainers.org/supported_docker_environment/continuous_integration/gitlab_ci/ for the how to use testcontainers-java in gitlab-ci

GregHilston:
I currently get a timeout

```
Pulling image mysql:5.7.28
⠏
Container started:  0c46b2b744
Waiting to be ready...
```

I'm using the `mysql` container and the exact log output is:

```
======================================================================= ERRORS =======================================================================
_________________________________________________________ ERROR at setup of test_with_mysql __________________________________________________________

    @pytest.fixture(scope="session")
    def mysql():
>       mysql = MySqlContainer("mysql:5.7.28").start()

tests/service/test_functional.py:9:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
venv/lib/python3.7/site-packages/testcontainers/core/generic.py:49: in start
    self._connect()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

[...]

Issue #25, opened 2018-11-26
Error running test inside a container


ori-n:
Hi,

I'm running my tests as part of the development pipeline inside a container. When I tried test containers I got the following error:
`requests.exceptions.ConnectionError: ('Connection aborted.', FileNotFoundError(2, 'No such file or directory'))`

Can someone please help?

Thanks

aleksandr-kotlyar:
Hi, have the same issue.

- **Scenario:**
1. build image from Dockefile below with the following requirements
2. run the container 
3. copy tests to container
4. execute pytest command for tests in container
- **Result:** tests fail with errors, with such a log, as below.
- **Expected result:** tests are running normally.

- **Requirements:**
```
pytest==4.1.1
requests==2.21.0
selene==1.0.0a13
testcontainers==2.3
allure-pytest==2.6.2
```

 - **Dockerfile:**
```
FROM python:3.6.6-alpine3.7

ADD requirements.txt /app/requirements.txt

WORKDIR /app

VOLUME /var/run/docker.sock:/var/run/docker.sock

RUN pip install --no-cache-dir -r requirements.txt
```

- **Detailed log:**

```
------------------------------ Captured log setup ------------------------------
conftest.py                 10 INFO     browser init
_ ERROR at setup of TestMainSliderControls.test_slider_arrows_both_active_on_the_middle_slides _
/usr/local/lib/python3.6/site-packages/urllib3/connectionpool.py:600: in urlopen
    chunked=chunked)
/usr/local/lib/python3.6/site-packages/urllib3/connectionpool.py:354: in _make_request
    conn.request(method, url, **httplib_request_kw)
/usr/local/lib/python3.6/http/client.py:1239: in request
    self._send_request(method, url, body, headers, encode_chunked)
/usr/local/lib/python3.6/http/client.py:1285: in _send_request
    self.endheaders(body, encode_chunked=encode_chunked)
/usr/local/lib/python3.6/http/client.py:1234: in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
/usr/local/lib/python3.6/http/client.py:1026: in _send_output
    self.send(msg)
/usr/local/lib/python3.6/http/client.py:964: in send
    self.connect()
/usr/local/lib/python3.6/site-packages/docker/transport/unixconn.py:43: in connect
    sock.connect(self.unix_socket)
E   FileNotFoundError: [Errno 2] No such file or directory

During handling of the above exception, another exception occurred:
[...]

PR #70, opened 2020-02-27
Add gitlab ci.

tillahoffmann:
There appear to be a number of issues with running tests on gitlab, e.g. #68, #60, #43, #25. Hopefully running tests on gitlab as part of the build will help us catch those issues.

Unfortunately, because of https://gitlab.com/gitlab-org/gitlab/issues/5667, we cannot run gitlab builds on pull requests from forks.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -10,4 +10,4 @@ tests/3.8:
   image: "python:3.8"
   script:
     - pip install -r requirements/3.8.txt
-    - pytest tests -v
+    - pytest tests -s -v --log-cli-level=debug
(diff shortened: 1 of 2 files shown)


SergeyPirogov:
Why not use Github's actions?

tillahoffmann:
@SergeyPirogov, updated above with rationale.

SergeyPirogov:
@tillahoffmann I see, good. So then it's better to provide a link to GitLab CI in the readme

tillahoffmann:
Yes, good idea. Will add a link to the readme, but will try to get it to build first. 🙂

tillahoffmann:
Available at https://github.com/testcontainers/testcontainers-python/blob/gitlab2/.gitlab-ci.yml if required, but [gitlab doesn't support fork builds](https://gitlab.com/gitlab-org/gitlab/-/issues/5667) which makes it problematic for open source use.

PR #78, opened 2020-03-23
Docker-compose --build support

### LINKED ISSUE #76: Option to build the docker containers before starting
When using this package to test a custom container, it'd be good to build the containers before starting the service every time the test runs to make sure any new changes to the Dockerfiles are tested.

I think this can be a param to the `DockerCompose` constructor:
`def __init__(self, filepath, compose_file_name="docker-compose.yml", build=False):`

If `True`, it would add a `--build` option to the `docker-compose` command inside `start()`

jtanios:
This PR adds support to build the containers before starting the services.
Resolves #76 

Also, there's an enhancement to the tests to make them run from anywhere, it doesn't have to be project root. This was an issue when running the tests from an IDE, like PyCharm.

### DIFF
--- a/tests/test_docker_compose.py
+++ b/tests/test_docker_compose.py
@@ -3,10 +3,11 @@ import pytest
 from testcontainers.compose import DockerCompose
 from testcontainers.core.exceptions import NoSuchPortExposed
 from testcontainers.core.utils import inside_container
+from tests import TESTS_DIR
 
 
 def test_can_spawn_service_via_compose():
-    with DockerCompose("tests") as compose:
+    with DockerCompose(TESTS_DIR) as compose:
         host = compose.get_service_host("hub", 4444)
         port = compose.get_service_port("hub", 4444)
         assert host == "0.0.0.0"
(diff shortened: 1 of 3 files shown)


jtanios:
@tillahoffmann , would you like to review?

tests/test_docker_compose.py
```diff
@@ -3,31 +3,40 @@
...
-    with DockerCompose("tests", pull=True) as compose:
+    with DockerCompose(TESTS_DIR, pull=True) as compose:
+        host = compose.get_service_host("hub", 4444)
+        port = compose.get_service_port("hub", 4444)
+        assert host == "0.0.0.0"
+        assert port == "4444"
+
+
+def test_can_build_images_before_spawning_service_via_compose():
+    with DockerCompose(TESTS_DIR, build=True) as compose:
```
tillahoffmann: What does `docker-compose` do when `--build` is passed but the [`docker-compose.yml`](https://github.com/testcontainers/testcontainers-python/blob/master/tests/docker-compose.yml) file doesn't contain any reference to a locally-built image?

jtanios: Good question!

I just tried it with the `tests/docker-compose.yml` and it had no effect; it worked exactly as it would without it.

jtanios: @tillahoffmann :point_up: is this good enough?

tillahoffmann:
Closing for now as `wontfix` (see #83 for details), but thank you for the contribution!

PR #84, opened 2020-04-21
Feature/sqlserver

annabaas:
SQL Server support, now using pyodbc. I expect this run to fail because pyodbc requires the client to install a driver.

### DIFF
--- a/testcontainers/mssql.py
+++ b/testcontainers/mssql.py
@@ -34,9 +34,9 @@ class SqlServerContainer(DbContainer):
 
     def get_connection_url(self):
         standard_url = super()._create_connection_url(dialect="mssql+pyodbc",
-                                              username=self.SQLSERVER_USER,
-                                              password=self.SQLSERVER_PASSWORD,
-                                              db_name=self.SQLSERVER_DBNAME,
-                                              port=self.port_to_expose)
+                                                      username=self.SQLSERVER_USER,
+                                                      password=self.SQLSERVER_PASSWORD,
+                                                      db_name=self.SQLSERVER_DBNAME,
+                                                      port=self.port_to_expose)
 
-        return standard_url + f"?driver={self.SQLSERVER_DRIVER}"
+        return standard_url + "?driver=" + self.SQLSERVER_DRIVER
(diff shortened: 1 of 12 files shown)


testcontainers/mssql.py
```diff
@@ -0,0 +1,42 @@
...
+
+class SqlServerContainer(DbContainer):
+    SQLSERVER_PASSWORD = environ.get("SQLSERVER_PASSWORD", "1Secure*Password1")
+    SQLSERVER_USER = "SA"
+    SQLSERVER_DBNAME = "tempdb"
+    SQLSERVER_DRIVER = "ODBC Driver 17 for SQL Server"
+
+    def __init__(self, image="mcr.microsoft.com/mssql/server:2019-latest", **kwargs):
+        super(SqlServerContainer, self).__init__(image)
+        self.port_to_expose = 1433
```
tillahoffmann, line 14: Can we add `port_to_expose` to the arguments in case someone wants to spin up two of these?

testcontainers/mssql.py
```diff
@@ -0,0 +1,42 @@
...
+        self.with_exposed_ports(self.port_to_expose)
+        self.ACCEPT_EULA = 'Y'
+        self.MSSQL_PID = 'Developer'
+
+        if 'SQLSERVER_PASSWORD' in kwargs:
```
tillahoffmann, line 19: Can we add these as explicit parameters to the `__init__` signature? It'll make the constructor easier to understand, especially for things like [`jedi`](https://github.com/davidhalter/jedi). E.g. something like

```python
self.SQLSERVER_PASSWORD = sqlserver_password or self.SQLSERVER_PASSWORD
```

where `sqlserver_password` is an argument of the constructor.

testcontainers/mssql.py
annabaas: Done!

testcontainers/mssql.py
annabaas: And done! Thank you for the lightning-fast review, I'd be glad to hear it if there's anything else!

tillahoffmann:
Thank you for the addition, @annabaas! I'll try to push a new release over the next couple of days to get things on pypi.

PR #91, opened 2020-05-06
docker-compose: Support input of Multiple Compose Files

Ragora:
At work, we are utilizing docker containers instrumented via docker-compose very heavily and need to be able to have fine control over what the docker-compose tool ingests when running tests. 

Even if this pull isn't accepted as-is, the ability to do this otherwise would be appreciated.


Ragora:
Handled by #98

Issue #92, opened 2020-05-25
Allow to fetch logs from containers run by compose


romank0:
I would like to fetch logs returned by `docker-compose logs` for containers run by `DockerCompose`.

thedrow:
Shouldn't this be closed since #93 was merged?

PR #93, opened 2020-05-25
Allow to get logs from compose

romank0:
This PR adds ability to get logs from containers started by `DockerCompose`. See https://github.com/testcontainers/testcontainers-python/issues/92

### DIFF
--- a/testcontainers/compose.py
+++ b/testcontainers/compose.py
@@ -30,6 +30,9 @@ class DockerCompose(object):
                 desired_capabilities=CHROME,
             )
             driver.get("http://automation-remarks.com")
+            stdout, stderr = compose.get_logs()
+            if stderr:
+                print("Errors\n:{}".format(stderr))
 
 
     .. code-block:: yaml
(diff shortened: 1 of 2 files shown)


testcontainers/compose.py
```diff
@@ -80,6 +83,16 @@ def stop(self):
             subprocess.call(["docker-compose", "-f", self.compose_file_name, "down", "-v"],
                             cwd=self.filepath)
 
+    def get_logs(self):
+        with blindspin.spinner():
```
tillahoffmann, line 87: Thoughts on using `subprocess.run` rather than the low-level `Popen` as recommended [here](https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module), i.e. something like this?

```python
result = subprocess.run(["docker-compose", "-f", self.compose_file_name, "logs"],
                        cwd=self.filepath, capture_output=True)
return result.stdout, result.stderr
```

romank0: It makes sense. Fixed.

testcontainers/compose.py
```diff
@@ -80,6 +83,16 @@ def stop(self):
...
+            result = subprocess.run(
+                ["docker-compose", "-f", self.compose_file_name, "logs"],
+                cwd=self.filepath,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
```
tillahoffmann, line 92:
```suggestion
                cwd=self.filepath, capture_output=True
```

romank0: @tillahoffmann This does not work as `capture_output` was only added in 3.7.  See this commit https://github.com/testcontainers/testcontainers-python/pull/93/commits/7fe7631a108e0e1e678bf6ceb74a4766a9c1e8e6 and failed tests on it.

romank0: Version that works in all supported version of python is in https://github.com/testcontainers/testcontainers-python/pull/93/commits/8a305e4af576a26f5b035e9937c09726f789262e

tillahoffmann: Duh, sorry, let me revert that. 🤦

PR #95, opened 2020-05-26
Update python_requires to have minimum of 3.5

### LINKED ISSUE #94: Installation with poetry on python > 3.5
Hi,

It is currently impossible to install this library using poetry that has python version requirements of > 3.5.

This is due to the `python_requires` in your `setup.py` of `'~=3.5',`: https://github.com/testcontainers/testcontainers-python/blob/master/setup.py#L64

Poetry fails with an error message like so:
```
[SolverProblemError]
The current project's Python requirement (>=3.7) is not compatible with some of the required packages Python requirement:
  - testcontainers requires Python ~=3.5

Because testcontainers (3.0.0) requires Python ~=3.5
 and no versions of testcontainers match >3.0.0,<4.0.0, testcontainers is forbidden.
So, because data-upload depends on testcontainers (^3.0.0), version solving failed.
```

Given that this library is tested with versions 3.5 -> 3.8 in travis, i think this python requirement is incorrect.

dlmiddlecote:
Closes #94 

Signed-off-by: Daniel Middlecote <dlmiddlecote@gmail.com>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,5 +61,5 @@ setuptools.setup(
     },
     long_description_content_type="text/x-rst",
     long_description=long_description,
-    python_requires='~=3.5',
+    python_requires='>=3.5',
 )
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ with open('README.rst') as fp:
 setuptools.setup(
     name='testcontainers',
     packages=setuptools.find_packages(exclude=['tests']),
-    version='3.0.1',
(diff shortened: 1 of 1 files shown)


tillahoffmann:
Thanks for the fix, @dlmiddlecote. Would you mind bumping the version to `3.0.2`? Then we can release a new version.

dlmiddlecote:
Updated @tillahoffmann. Thanks for the quick reply!

PR #96, opened 2020-05-29
Fallback to gateway address if it is running inside a container

tuvshuud:
Fix:
- Stuck when running inside docker container with docker socket is mounted. Reported by @karmux [here](https://github.com/testcontainers/testcontainers-python/issues/43#issuecomment-635504867) in issue #43


tillahoffmann:
@tuvshuud, thank you for this. Would you mind bumping the patch version in this PR and we can cut a new release?

tuvshuud:
@tillahoffmann I don’t mind. This is okay to be in patch version.

Issue #99, opened 2020-06-07
MongoDbContainer on Windows 10 fails with getaddrinfo error


arpost:
The following code fails with testcontainers-python 3.0.2:
```
from testcontainers.mongodb import MongoDbContainer
with MongoDbContainer('mongo:4.2.2') as mongo_:
...     db = mongo_.get_connection_client().test
...     db.test.insert_one({'foo': 'bar'})
```
with
```
Pulling image mongo:4.2.2
Container started:  d5ca66c443
Traceback (most recent call last):
  File "<input>", line 3, in <module>
  File "C:\...\venv\lib\site-packages\pymongo\collection.py", line 695, in insert_one
    self._insert(document,
  File "C:\...\venv\lib\site-packages\pymongo\collection.py", line 610, in _insert
    return self._insert_one(
  File "C:\...\venv\lib\site-packages\pymongo\collection.py", line 599, in _insert_one
    self.__database.client._retryable_write(
  File "C:\...\venv\lib\site-packages\pymongo\mongo_client.py", line 1490, in _retryable_write
    with self._tmp_session(session) as s:
  File "C:\Program Files\Python38\lib\contextlib.py", line 113, in __enter__
    return next(self.gen)
  File "C:\...\venv\lib\site-packages\pymongo\mongo_client.py", line 1823, in _tmp_session
    s = self._ensure_session(session)
  File "C:\...\venv\lib\site-packages\pymongo\mongo_client.py", line 1810, in _ensure_session
    return self.__start_session(True, causal_consistency=False)
  File "C:\...\venv\lib\site-packages\pymongo\mongo_client.py", line 1763, in __start_session
    server_session = self._get_server_session()
  File "C:\...\venv\lib\site-packages\pymongo\mongo_client.py", line 1796, in _get_server_session
    return self._topology.get_server_session()
  File "C:\...\venv\lib\site-packages\pymongo\topology.py", line 482, in get_server_session
    self._select_servers_loop(
  File "C:\...\venv\lib\site-packages\pymongo\topology.py", line 208, in _select_servers_loop
    raise ServerSelectionTimeoutError(
pymongo.errors.ServerSelectionTimeoutError: localnpipe:32775: [Errno 11001] getaddrinfo failed
```

Interestingly, the `MongoDbContainer.get_connection_url()` method returns a URL with hostname localnpipe, which I suspect the mongo client does not understand. Using pymongo directly, I can connect to the created container using a hostname of localhost.

[...]

PR #100, opened 2020-06-08
Fix for getaddrinfo failures on Windows 10 in MongoDbContainer

arpost:
Windows npipe URIs are rewritten by the docker-py client library to http+docker://localnpipe. Other pipe-like URIs are rewritten this way too. So, `core/docker_client.py` should detect the http+docker scheme instead of "npipe" and return localhost. Because the check for "http" schemes will inadvertently pick up the http+docker scheme too, do the check for pipes first and for real URLs second. Fixes issue #99. All tests pass with this fix on Windows 10 with Docker Desktop running locally.

### DIFF
--- a/testcontainers/core/docker_client.py
+++ b/testcontainers/core/docker_client.py
@@ -62,7 +62,7 @@ class DockerClient(object):
 
         except ValueError:
             return None
-        if 'unix' in url.scheme or 'http+docker' == url.scheme:
+        if 'unix' in url.scheme or 'npipe' in url.scheme or 'http+docker' == url.scheme:
             if inside_container():
                 ip_address = default_gateway_ip()
                 if ip_address:
--- a/testcontainers/core/docker_client.py
+++ b/testcontainers/core/docker_client.py
@@ -62,11 +62,12 @@ class DockerClient(object):
 
         except ValueError:
             return None
-        if 'http' in url.scheme or 'tcp' in url.scheme:
-            return url.hostname
-        if 'unix' in url.scheme or 'npipe' in url.scheme:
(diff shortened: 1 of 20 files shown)


testcontainers/core/docker_client.py
```diff
@@ -62,11 +62,12 @@ def host(self):
 
         except ValueError:
             return None
-        if 'http' in url.scheme or 'tcp' in url.scheme:
-            return url.hostname
-        if 'unix' in url.scheme or 'npipe' in url.scheme:
+        if 'unix' in url.scheme or 'http+docker' == url.scheme:
```
tillahoffmann, line 65: Should `npipe` also be included in this check for backwards compatibility?

arpost:
That’s a good idea. I don’t know when the docker client started replacing npipe with the special URL.

arpost:
I updated the pull request with the npipe check.

tests/test_nginx_container.py
```diff
@@ -9,6 +10,7 @@ def test_docker_run_nginx():
         port = nginx.port_to_expose
         url = "http://{}:{}/".format(nginx.get_container_host_ip(),
                                      nginx.get_exposed_port(port))
+        time.sleep(5)
```
tillahoffmann, line 13: Can we drop the sleeps?

arpost:
I have those in my copy because the tests fail on my hardware without them, but I didn’t mean for them to end up in my pull request. Sorry about that, I’ll move them so they aren’t in the pull request.

alexanderankin:
what was the end outcome here?

PR #103, opened 2020-06-26
Fixed MongoDbContainer to actually use username and password

fsonntag:
First thanks for the library :)

I figured that the MongoDB container doesn't actually configure the username and password, so I added that it works


tillahoffmann:
Hey, thank you for the fix! Would you be able to add a test as well (just to make sure we don't break things again in the future)?

fsonntag:
@tillahoffmann Thanks for the feedback! I think the most important thing was that the existing test keeps working. Anyway, I now added another test for connecting without authentication enabled. Let me know if you think another kind of test is required.

PR #107, opened 2020-07-03
Add Localstack container support

### LINKED ISSUE #33: Feature Request:  Localstack container
Localstack is a mock of AWS services.  There is a Localstack available in the testcontainers-java library: https://github.com/testcontainers/testcontainers-java/blob/master/modules/localstack/src/main/java/org/testcontainers/containers/localstack/LocalStackContainer.java

It would be very useful if this were directly available in the python module.  My teams current workaround is to use the DockerCompose object with this docker-compose.yaml file:

```
version: '2.1'

services:
  localstack:
    image: localstack/localstack
    ports:
      - "4567-4597:4567-4597"
      - "${PORT_WEB_UI-8080}:${PORT_WEB_UI-8080}"
    environment:
      - SERVICES="kinesis,lambda"
      - DEBUG=true
      - USE_SSL=true
      - DOCKER_HOST=unix:///var/run/docker.sock
```

ImFlog:
This PR adds an equivalent to the testcontainers-java Localstack container.
This is simpler than the Java version because a simple edge port was introduced in version 0.11.0. 
This should fix #33.

### DIFF
--- a/testcontainers/localstack.py
+++ b/testcontainers/localstack.py
@@ -47,7 +47,7 @@ class LocalStackContainer(DockerContainer):
     def get_endpoint_override(self):
         host = self.get_container_host_ip()
         port = self.get_exposed_port(LocalStackContainer.EDGE_PORT)
-        return f'http://{host}:{port}'
+        return 'http://{}:{}'.format(host, port)
 
     def start(self):
         super().start()
--- a/testcontainers/localstack.py
+++ b/testcontainers/localstack.py
@@ -1,0 +1,58 @@
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+from testcontainers.core.waiting_utils import wait_for_logs
+from testcontainers.core.container import DockerContainer
+
+
+class LocalStackContainer(DockerContainer):
+    """
+    Localstack container.
+
+    Example
+    -------
+    ::
+        localstack = LocalStackContainer(image="localstack/localstack:0.11.3")
+        localstack.with_services("dynamodb", "lambda")
+        localstack.start()
+        dynamo_endpoint = localstack.get_service_endpoint()
+        dynamo_client = boto3.client("dynamodb", endpoint_url=dynamo_endpoint)
+        scan_result = dynamo_client.scan(TableName='foo')
+        # Do something with the scan result
+    """
+    EDGE_PORT = 4566
+    IMAGE = 'localstack/localstack:0.11.3'
+
+    def __init__(self, image=IMAGE):
+        super(LocalStackContainer, self).__init__(image)
+        self.with_exposed_ports(LocalStackContainer.EDGE_PORT)
+        self.with_volume_mapping('/var/run/docker.sock', '/var/run/docker.sock')
+
+    def with_services(self, *services):
+        """
+        Restrict what services to run.
+        If you don't call this, everything will be launched by default.
+        """
+        self.with_env('SERVICES', ','.join(services))
+
+    def get_endpoint_override(self):
+        host = self.get_container_host_ip()
+        port = self.get_exposed_port(LocalStackContainer.EDGE_PORT)
+        return f'http://{host}:{port}'
+
+    def start(self):
+        super().start()
+        self._get_ready_log()
+        return self
+
+    def _get_ready_log(self, timeout=30):
+        wait_for_logs(self, 'Ready\\.', timeout=timeout)
--- a/testcontainers/localstack.py
+++ b/testcontainers/localstack.py
@@ -24,7 +24,7 @@ class LocalStackContainer(DockerContainer):
         localstack = LocalStackContainer(image="localstack/localstack:0.11.3")
         localstack.with_services("dynamodb", "lambda")
         localstack.start()
-        dynamo_endpoint = localstack.get_service_endpoint()
+        dynamo_endpoint = localstack.get_endpoint_override()
         dynamo_client = boto3.client("dynamodb", endpoint_url=dynamo_endpoint)
         scan_result = dynamo_client.scan(TableName='foo')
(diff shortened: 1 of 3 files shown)


testcontainers/localstack.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+    Localstack container.
+
+    Example
+    -------
+    ::
+        localstack = LocalStackContainer(image="localstack/localstack:0.11.3")
+        localstack.with_services("dynamodb", "lambda")
+        localstack.start()
+        dynamo_endpoint = localstack.get_service_endpoint()
```
hermanschaaf, line 27:
```suggestion
        dynamo_endpoint = localstack.get_endpoint_override()
```

When running this locally, I found `get_service_endpoint` to not be defined. Should this be `get_endpoint_override`?

testcontainers/localstack.py
```diff
@@ -0,0 +1,58 @@
...
+        host = self.get_container_host_ip()
+        port = self.get_exposed_port(LocalStackContainer.EDGE_PORT)
+        return 'http://{}:{}'.format(host, port)
+
+    def start(self):
+        super().start()
+        self._get_ready_log()
+        return self
+
+    def _get_ready_log(self, timeout=30):
```
hermanschaaf, line 57:
```suggestion
    def _get_ready_log(self, timeout=60):
```

Using this with only the S3 service often takes longer than 30 seconds, at least on my machine. 60 seconds seems to be enough.

It might also be nice to be able to override this timeout as a parameter to the `start` method

testcontainers/localstack.py
```diff
@@ -0,0 +1,58 @@
...
+        scan_result = dynamo_client.scan(TableName='foo')
+        # Do something with the scan result
+    """
+    EDGE_PORT = 4566
+    IMAGE = 'localstack/localstack:0.11.3'
+
+    def __init__(self, image=IMAGE):
+        super(LocalStackContainer, self).__init__(image)
+        self.with_exposed_ports(LocalStackContainer.EDGE_PORT)
+        self.with_volume_mapping('/var/run/docker.sock', '/var/run/docker.sock')
```
hermanschaaf, line 38: Is this line necessary? Does localstack spin up new containers in some circumstances?

ImFlog: For lambdas it needs to spin up containers yes. Most of the time I use this so I added it.
I can remove it or create a function doing this, WDYT ?

hermanschaaf: Ah, interesting. Would it be an idea to only set the volume mapping if the `lambda` service is requested? In some CI environments (such as ours) it would not be desirable for additional containers to be created, as they might not get cleaned up afterwards

ImFlog: I am not sure that only lambda needs this to be honest. Also some other ENV vars can change this behavior and make the lambda run locally instead of new containers.
So for now I think it's better if we let the users chose when to put this volume mapping according to what's span.
I will remove it.

testcontainers/localstack.py
```diff
@@ -0,0 +1,57 @@
...
+        port = self.get_exposed_port(LocalStackContainer.EDGE_PORT)
+        return 'http://{}:{}'.format(host, port)
+
+    def start(self, timeout=60):
+        super().start()
+        self._get_ready_log(timeout)
+        return self
+
+    def _get_ready_log(self, timeout):
+        wait_for_logs(self, 'Ready\\.', timeout=timeout)
```
hermanschaaf, line 57: Out of curiosity: why is there a slash here? (As if the string should be `Ready\.` ?)

ImFlog: Without it PEP8 was yelling so I did this.
Your comment made me realize that indeed this was strange and now It's clearer :)

ImFlog:
The comment were super useful, thanks @hermanschaaf !
Now let's wait for a maintainer review (I think it's @tillahoffmann).

ImFlog:
Anyone having the right on the repository willing to review this ?
AFAIK there is at least 1 company (mine) and @hermanschaaf that are interested in this :)

Iprogrammerr:
My company is interested also. We have a lot of small python apps that interact with aws environment. We are already using testcontainers for our db tests and it's great :)

testcontainers/localstack.py
```diff
@@ -0,0 +1,57 @@
...
+    def __init__(self, image=IMAGE):
+        super(LocalStackContainer, self).__init__(image)
+        self.with_exposed_ports(LocalStackContainer.EDGE_PORT)
+
+    def with_services(self, *services):
+        """
+        Restrict what services to run.
+        If you don't call this, everything will be launched by default.
+        """
+        self.with_env('SERVICES', ','.join(services))
```
tillahoffmann, line 44:
```suggestion
        return self.with_env('SERVICES', ','.join(services))
```

Small tweak to allow chaining of `with_*` calls.

testcontainers/localstack.py
```diff
@@ -0,0 +1,57 @@
...
+        host = self.get_container_host_ip()
+        port = self.get_exposed_port(LocalStackContainer.EDGE_PORT)
+        return 'http://{}:{}'.format(host, port)
+
+    def start(self, timeout=60):
+        super().start()
+        self._get_ready_log(timeout)
+        return self
+
+    def _get_ready_log(self, timeout):
```
tillahoffmann, line 56: Is this function being used elsewhere? If not, let's inline it within `start`.

testcontainers/localstack.py
```diff
@@ -0,0 +1,57 @@
...
+        self.with_exposed_ports(LocalStackContainer.EDGE_PORT)
+
+    def with_services(self, *services):
+        """
+        Restrict what services to run.
+        If you don't call this, everything will be launched by default.
+        """
+        self.with_env('SERVICES', ','.join(services))
+
+    def get_endpoint_override(self):
```
tillahoffmann, line 46: Can we add a docstring here? Do you think it makes sense to use the same naming (`get_url`) as in the `ElasticSearchContainer`?

https://github.com/testcontainers/testcontainers-python/blob/56e6b88d151509c83487f593efa8b12654a47673/testcontainers/elasticsearch.py#L43-L46

ImFlog: In the boto3 library, when overriding the url of a client a kwarg `endpoint_url=XXX` is used.
Maybe I can rename to that so it easier to understand when using the testcontainer ?
I can also change to get_url if you want.

tillahoffmann: Yes, let's go with `get_url` for consistency.

shashanksingh:
Heya, my org would also love to use it, we are waiting for this merge to write e2e tests. Please do tell me if I can help anyway.

tillahoffmann (approved):

PR #115, opened 2020-09-11
Fix LocalStackContainer class doc example

GFernie:


testcontainers/localstack.py
```diff
@@ -24,7 +24,7 @@ class LocalStackContainer(DockerContainer):
         localstack = LocalStackContainer(image="localstack/localstack:0.11.3")
```
GFernie, line 24: This is an older version. Update to latest? Remove image kwarg and use default?

```suggestion
        localstack = LocalStackContainer(image="localstack/localstack:0.11.4")
```

tillahoffmann (approved):

PR #120, opened 2020-10-29
Add GitHub actions

### LINKED ISSUE #71: Move from Travis to Github Actions
Time to move to native Github actions

maxim-nazarenko:
Closes #71

This PR adds 2 GitHub workflows:
1. `main` - regular push events
1. `pypi-release` - releasing the tagged version to PyPi

The release workflow is triggered by pushing any tag like `v*` but GitHub actions support [release events](https://docs.github.com/en/free-pro-team@latest/actions/guides/building-and-testing-python#publishing-to-package-registries) which might be considered as well.

Please, let me know about your thoughts regarding this PR so I could adjust its logic.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,5 +1,6 @@
 name: testcontainers-python
-on: [push]
+on:
+  - push
 
 jobs:
   build:
@@ -37,10 +38,3 @@ jobs:
           sphinx-build -nW docs docs/_build/html
           py.test -sv --cov-config .coveragerc --cov-report html:skip-covered --cov-report term:skip-covered --cov=testcontainers --tb=short tests/
           codecov
-  deploy:
-    needs: build
-    if: needs.build.result == 'success'
-    runs-on: ubuntu-16.04
-    steps:
-      - name: Deploy to PyPi
-        run: echo deploying
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,6 +1,9 @@
 name: testcontainers-python
 on:
-  - push
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
 
 jobs:
   build:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       matrix:
         python-version: [3.5, 3.6, 3.7, 3.8]
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-18.04
     steps:
       - uses: actions/checkout@v2
       - name: Setup python ${{ matrix.python-version }}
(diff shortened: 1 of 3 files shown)


tillahoffmann:
This is great, thank you. Apologies for the slow reply. 

Would you mind deleting the code that pushes the pypi package as part of the Travis build so we don't push twice?

maxim-nazarenko:
thanks for the reply @tillahoffmann,
I've updated the PR.
Looking forward for your feedback.

.github/workflows/main.yml
```diff
@@ -0,0 +1,40 @@
...
+jobs:
+  build:
+    strategy:
+      matrix:
+        python-version: [3.5, 3.6, 3.7, 3.8]
+    runs-on: ubuntu-16.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
```
tillahoffmann, line 14: We need to supply the version information here for the setup script to pick things up. Cf. https://github.com/tillahoffmann/idxhound/blob/master/.github/workflows/pythonpackage.yml#L25-L26

.github/workflows/pypi-release.yml
```diff
@@ -0,0 +1,39 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-16.04
+    env:
+      python-version: 3.8
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Setup python ${{ env.python-version }}
+        uses: actions/setup-python@v2
```
tillahoffmann, line 16: Same comment re python version here.

tillahoffmann (commented):
Thanks for the update. This doesn't run for me locally atm (using `act -P ubuntu-latest=nektos/act-environments-ubuntu:18.04 -v`). Comments inline.

.github/workflows/main.yml
maxim-nazarenko: thanks for pointing that out. I can't believe I missed that.

Updated the code.

.github/workflows/pypi-release.yml
maxim-nazarenko: done

maxim-nazarenko:
@tillahoffmann I haven't used the `act` command. Let me check it locally as well. Probably, it could be due to using `ubuntu-18` (in your example) instead of `ubuntu-16` (from actions YAML file)

.github/workflows/main.yml
```diff
@@ -0,0 +1,42 @@
+name: testcontainers-python
+on:
+  - push
```
tillahoffmann, line 3: Do we need something for pull requests here, e.g. something like https://github.com/tillahoffmann/localscope/blob/1584b66ce2e1cd02aa10b82332b5775f6ebcbdb7/.github/workflows/build.yml#L6-L12?

Or is that already covered?

.github/workflows/main.yml
```diff
@@ -0,0 +1,42 @@
...
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        python-version: [3.5, 3.6, 3.7, 3.8]
+    runs-on: ubuntu-16.04
```
tillahoffmann, line 10: Let's update to a more recent version of ubuntu. 16.04 is end of life in a few months.

tillahoffmann (commented):
Thanks for pointing out the os version inconsistency. That did indeed fix it. Few more comments.

.github/workflows/main.yml
maxim-nazarenko: done

.github/workflows/main.yml
maxim-nazarenko: no, it wasn't covered. I did a mirror translation from TravisCI file to Github action file.

Added `pull_request` trigger action.

PR #127, opened 2021-01-05
adding kafka test container support

### LINKED ISSUE #121: How to create a Kafka test Container
Could we create a Kafka container like this 
`with TestContainer(image="confluentinc/cp-kafka:5.4.3") as container:
    ip = container.get_container_host_ip()`

but I don't know how to implement a method like https://www.testcontainers.org/modules/kafka/
`kafka.getBootstrapServers()`

ash1425:
fix #121 This PR adds support for kafka container. Heavily inspired from java testcontainer's Kafka module.


ash1425:
I tried these failing tests locally, They work in all python versions (used venv) but were flaky and I had to try multiple time. 
mssql test didnt work locally too.

SrinivasanTarget:
@tillahoffmann Can you please review this change?

testcontainers/kafka.py
```diff
@@ -0,0 +1,73 @@
...
+                      'BROKER:PLAINTEXT,PLAINTEXT:PLAINTEXT')
+        self.with_env('KAFKA_INTER_BROKER_LISTENER_NAME', 'BROKER')
+
+        self.with_env('KAFKA_BROKER_ID', '1')
+        self.with_env('KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR', '1')
+        self.with_env('KAFKA_OFFSETS_TOPIC_NUM_PARTITIONS', '1')
+        self.with_env('KAFKA_LOG_FLUSH_INTERVAL_MESSAGES', '10000000')
+        self.with_env('KAFKA_GROUP_INITIAL_REBALANCE_DELAY_MS', '0')
+
+    def get_bootstrap_servers(self):
```
tillahoffmann, line 30: Should this be `get_bootstrap_server` (without s) as it only returns one value?

testcontainers/kafka.py
```diff
@@ -0,0 +1,73 @@
...
+        port = self.get_exposed_port(KafkaContainer.KAFKA_PORT)
+        return '{}:{}'.format(host, port)
+
+    @wait_container_is_ready()
+    def _connect(self):
+        bootstrap_servers = self.get_bootstrap_servers()
+        consumer = kafka.KafkaConsumer(group_id='test', bootstrap_servers=[bootstrap_servers])
+        topics = consumer.topics()
+        if not topics:
+            raise Exception()
```
tillahoffmann, line 41: Can we raise a more specific exception here (including an error message)?

testcontainers/kafka.py
```diff
@@ -0,0 +1,73 @@
...
+
+        self.with_env('KAFKA_BROKER_ID', '1')
+        self.with_env('KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR', '1')
+        self.with_env('KAFKA_OFFSETS_TOPIC_NUM_PARTITIONS', '1')
+        self.with_env('KAFKA_LOG_FLUSH_INTERVAL_MESSAGES', '10000000')
+        self.with_env('KAFKA_GROUP_INITIAL_REBALANCE_DELAY_MS', '0')
+
+    def get_bootstrap_servers(self):
+        host = self.get_container_host_ip()
+        port = self.get_exposed_port(KafkaContainer.KAFKA_PORT)
```
tillahoffmann, line 32: Should this be `self.port_to_expose`?

tests/test_kafka.py
```diff
@@ -0,0 +1,10 @@
+import kafka
+
+from testcontainers.kafka import KafkaContainer
+
+
+def test_docker_run_kafka():
+    with KafkaContainer() as container:
```
tillahoffmann, line 7: Can we try this with at least two different ports to make sure they all line up?

tillahoffmann (commented):
Thank you for the addition! Some comments inline.

testcontainers/kafka.py
ash1425: Kafka library doesn't have any specific exception. But we should definitely add the helpful message here.  👍🏼

tillahoffmann: I was thinking we should raise an exception here that is more specific than `Exception`, e.g. a `RuntimeError` or `ValueError` (or another one) depending on which is most appropriate.

tests/test_kafka.py
mamachanko: I would like to suggest to also test publishing and consuming messages.

ash1425: @tillahoffmann Could you please elaborate what do you mean by two different ports here?

ash1425:
@tillahoffmann Thanks for the review comments. I have accommodated those in latest commit.

tests/test_kafka.py
```diff
@@ -0,0 +1,32 @@
...
+    producer = KafkaProducer(bootstrap_servers=[bootstrap_server])
+    producer.send(topic, b"verification message")
+    producer.flush()
+    producer.close()
+
+    consumer = KafkaConsumer(topic, bootstrap_servers=[bootstrap_server])
+    consumer.partitions_for_topic(topic)
+    consumer.seek_to_beginning()
+
+    for msg in consumer:
```
mamachanko, line 30: This test passes if there are no messages. I would suggest to assert that it's non-empty in addition or an assertion like "assert [m for m in consumer] == [expected_msg]".

testcontainers/kafka.py
```diff
@@ -0,0 +1,74 @@
...
+echo 'dataLogDir=/var/lib/zookeeper/log' >> zookeeper.properties
+zookeeper-server-start zookeeper.properties &
+export KAFKA_ZOOKEEPER_CONNECT='localhost:2181'
+export KAFKA_ADVERTISED_LISTENERS=PLAINTEXT://localhost:{},BROKER://$(hostname -i):9092
+. /etc/confluent/docker/bash-config
+/etc/confluent/docker/configure
+/etc/confluent/docker/launch""".format(port).encode('utf-8')
+
+        tar_stream = BytesIO()
+        tar = tarfile.TarFile(fileobj=tar_stream, mode='w')
```
mamachanko, line 58: This could be improved through the use of context managers. Consider https://github.com/mamachanko/website-monitor/blob/main/tests/kafka.py#L84-L91 which I based on your code.

tests/test_kafka.py
ash1425: Using the code you suggested makes my test to wait indefinitely and eventually fail after some time. I have decided to use offset to verify that there is one message present on the topic before checking the contents.

testcontainers/kafka.py
```diff
@@ -0,0 +1,84 @@
...
+                echo 'clientPort=2181' > zookeeper.properties
+                echo 'dataDir=/var/lib/zookeeper/data' >> zookeeper.properties
+                echo 'dataLogDir=/var/lib/zookeeper/log' >> zookeeper.properties
+                zookeeper-server-start zookeeper.properties &
+                export KAFKA_ZOOKEEPER_CONNECT='localhost:2181'
+                export KAFKA_ADVERTISED_LISTENERS={listeners}
+                . /etc/confluent/docker/bash-config
+                /etc/confluent/docker/configure
+                /etc/confluent/docker/launch
+                """
```
mamachanko, line 61: Formatted string literals aka f-strings have been [introduced in Python 3.6](https://docs.python.org/3/whatsnew/3.6.html#whatsnew36-pep498). That's why the [tests for 3.5](https://travis-ci.org/github/testcontainers/testcontainers-python/jobs/756157684) fail when flake8 tries to parse this file.

Sorry, I might've pointed you in that direction with my example code. Apologies.

Seems like plain old `"".format()` is the way to go:
```python
"""#!/usr/bin/env bash

curl https://{domain}.com
""".format(domain="example")
```

mamachanko (changes requested):

testcontainers/kafka.py
ash1425: Thanks. I did encounter it before but forgot yesterday. Fixing now.

tests/test_kafka.py
mamachanko: I see your point. However, the test as it is may still pass with the messages in the consumer being `[]` because the assertion within the `for` loop would never be executed. That's not ideal.

Alternatively, you could consider to use `poll` (see the [docs](https://kafka-python.readthedocs.io/en/master/apidoc/KafkaConsumer.html#kafka.KafkaConsumer.poll).

The `break` statement isn't needed.

mamachanko (changes requested):

ash1425:
`assert consumer.end_offsets([tp])[tp] == 1, "Expected exactly one test message to be present on test topic !"`
        
This is going to ensure that `[]` is not accepted and we should have one and only one message in the consumer.

mamachanko:
> `assert consumer.end_offsets([tp])[tp] == 1, "Expected exactly one test message to be present on test topic !"`
> 
> This is going to ensure that we will have one and only one message in the consumer in integration test.

I understand. Then it looks like the `for` loop and the assertion within don't add confidence to the test. Would you consider to scrap them?

mamachanko (approved):

YikSanChan:
@ash1425 Hey Ashay, do you have bandwidth working on the PR? If not, maybe I can help to cross the finish line?

ash1425:
@YikSanChan 
Thanks for offering to help. I have invited you as collaborator to my fork.

For now, I have incorporated all the review comments and merged upstream changes to resolve the conflicts. I think we should be good to approve and merge if all tests pass.

YikSanChan:
@ash1425 Great thanks! If there is any more comment from testcontainers members, I can try to resolve

YikSanChan:
@tillahoffmann Could you please taking a look? This is the feature I am looking for. Thank you!

SrinivasanTarget:
@tillahoffmann CI seems happy. Are we good to get these changes in?

tillahoffmann (approved):

PR #128, opened 2021-01-21
[RedisContainer] Support decode_responses setting

ysde:
Support decode_responses setting

### DIFF
--- a/testcontainers/redis.py
+++ b/testcontainers/redis.py
@@ -17,10 +17,11 @@ from testcontainers.core.waiting_utils import wait_container_is_ready
 
 
 class RedisContainer(DockerContainer):
-    def __init__(self, image="redis:latest", port_to_expose=6379):
+    def __init__(self, image="redis:latest", port_to_expose=6379, decode_responses=False):
         super(RedisContainer, self).__init__(image)
         self.port_to_expose = port_to_expose
         self.with_exposed_ports(self.port_to_expose)
+        self.decode_responses = decode_responses
 
     @wait_container_is_ready()
     def _connect(self):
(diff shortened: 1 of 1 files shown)


ysde:
Hi @tillahoffmann 

Can you help confirm this pr ?

Thank you

testcontainers/redis.py
```diff
@@ -29,7 +30,7 @@ def _connect(self):
             raise Exception
 
     def get_client(self):
```
tillahoffmann, line 32: What are your thoughts on instead supporting keyword arguments for the `get_client` function?

ysde: Hi @tillahoffmann 

Thanks for the review.

I think it would be better in the constructor. Because it's the same behavior when we use the python [redis](https://pypi.org/project/redis/) library. So in the PR, I put it there in the first place.

```
def __init__(self, image="redis:latest", port_to_expose=6379, decode_responses=False)
```

`get_client` is also fine to me too.

I think it's up to you :)

Thank you

tillahoffmann: Okay, let's go with `**kwargs` on the `get_client` method because it allows different clients to be instantiated with different parameters. You might also be able to add `functools.wraps` as a decorator to provide a nicer docstring.

ysde: Hi @tillahoffmann 

I'm not familiar with `functors.wraps` and not sure how to add it on `get_client`, could you give me some examples ?

And there's a new commit can you help to see is it what you imagined ?

Thank you very much

tillahoffmann (commented):
Few more small edits.

ysde:
Hi @tillahoffmann 

I've done the modifications in the latest commit. : )

Thank you

ysde:
@tillahoffmann 

Thank you!

PR #136, opened 2021-04-16
General maintenance and tidying to ensure tests pass.

tillahoffmann:
This large-ish PR aims to fix a number of challenges that make some of the tests flaky. At a high level:

- Move from `pyodbc` to `pymssql` because the former hasn't proved particularly reliable and also doesn't work on `arm` platforms.
- Updated requirements that do not require code changes. Cf. #161.
- Change `wait_container_is_ready` to accept a list of "transient errors" for which another try should be attempted. Other errors (such as `ImportError`) reraise the exception immediately. This should avoid some confusion such as #137.
- Add a function `maybe_emulate_amd64` which sets `--platform=linux/amd64` if the platform is `arm64`. This is required for cross-architecture testing because some images are only available as `amd64`.
- Refactor of some parts of the docker client to provide easier error messages, such as "port mapping isn't available" as supposed to "NoneType is not indexable".
- Changed the script of the Kafka container from `sh` to `bash` because `sh` doesn't support some of the syntax used (namely brackets). The Kafka tests are still a bit unstable, and I've marked them as `xfail`.
- Some refactor in for MySql and SqlServer containers which used a mix of class variables and instance variables which required some hacks (e.g. using `reload` in testing).
- Added some branching to allow for the tests to run on both `arm64` and `amd64`.

@KerstenBreuer, this might be of interest if you want to get involved more with the code base.
@SergeyPirogov, would be great to get your eyes on this.


jcraftsman (approved):

tillahoffmann:
@yakimka, planning to merge this one soon as it's been lingering for almost a year. It's a bit of a beast (see description) above.

yakimka:
@tillahoffmann any suggestions on my comments?

tillahoffmann:
@yakimka, I can't see any comments. Are they possibly still showing as pending on your end?

yakimka:
@tillahoffmann oh, sorry, my fault

mihaylovich:
Hello.
I think logging is enough to quickly investigate the bug described by #137.
I mean this added code
` logger.debug('container is not yet ready: %s', traceback.format_exc())`

But add transient_exceptions tuple is confusing when use some other docker images.
e.g.  MS SQL container on certain state of starting server returns the error:
`pyodbc.InterfaceError: ('28000', "[28000] [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]Login failed for user 'SA'. (18456) (SQLDriverConnect); [28000] [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]Login failed for user 'SA'. (18456)")
`
These changes have also led to the need to list all the required exception types in the library's source code (e.g kafka.py)
When you instantiate any DockerContainer you can specify the concrete image name, but you cannot specify which exceptions should be ignored.

tillahoffmann:
Hey, we can definitely consider adding transient exceptions as an exposed argument. The motivation behind the change was to not "hide" errors when they occur. E.g. an `ImportError` would previously time out after two minutes even though it should've failed immediately.

PR #151, opened 2021-07-15
Add compose exec_in_container method

kieranlea:
This PR adds a method that reaches out to docker-compose to execute a command in a service container.

~~I'm open to suggestions, in particular on the return type. Having a 3 tuple undocumented seems like it may be unintuitive to understand, but there seems to already be precedent in the `get_logs` method above, and not seeing method docstrings there to document it.~~

This was an old PR and I see there is some precedence for docstrings now, added a docstring as well

Thanks

### DIFF
--- a/testcontainers/compose.py
+++ b/testcontainers/compose.py
@@ -115,7 +115,7 @@ class DockerCompose(object):
         Executes a command in the container of one of the services.
 
         Parameters
-        -------
+        ----------
         service_name: str
             Name of the docker compose service to run the command in
         command: list[str]
--- a/testcontainers/compose.py
+++ b/testcontainers/compose.py
@@ -110,6 +110,31 @@ class DockerCompose(object):
         )
         return result.stdout, result.stderr
 
+    def exec_in_container(self, service_name, command):
+        """
+        Executes a command in the container of one of the services.
+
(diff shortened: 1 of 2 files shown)


yakimka (approved):

yakimka:
@tillahoffmann, what do you think? Can we merge this?

tillahoffmann (changes requested):
Let's add some documentation. I appreciate that some of the methods are poorly documented, but we should change that, too.

kieranlea:
> Let's add some documentation. I appreciate that some of the methods are poorly documented, but we should change that, too.

Agreed, thanks for reviewing the PR

testcontainers/compose.py
```diff
@@ -110,6 +110,31 @@ def get_logs(self):
         )
         return result.stdout, result.stderr
 
+    def exec_in_container(self, service_name, command):
+        """
+        Executes a command in the container of one of the services.
+
+        Parameters
+        -------
```
tillahoffmann, line 118:
```suggestion
        ----------
```

tillahoffmann (approved):

tillahoffmann:
Merged after unrelated test failure.

PR #154, opened 2021-09-08
WIP: Host ip fix

gel-bcegielka:


gel-bcegielka:
Apologies, I was meant to create PR in my fork, but didn't change the base. This can be closed as it was a dirty hack to get testcontainers in gitlab dind setup.
It such case
https://github.com/testcontainers/testcontainers-python/blob/master/testcontainers/core/container.py#L79
function `get_container_host_ip` returns container’s bridge ip 178.18.0.1
`Connection url: postgresql+psycopg2://test:test@172.18.0.1:32768/test`
which does not work.

What we need it docker service we started, which is
$ ping -c 1 docker
PING docker (172.17.0.2): 56 data bytes

PR #156, opened 2021-09-25
Support specifying postgresql drivers

KarolisL:
I've found installing `psycopg2` driver on MacOS Big Sur being too complicated: I didn't want to install `postgres` headers nor was I able to use the `-binary` version of the package. 

I've decided to use pure-python driver, but wasn't able to do so due to the driver being hardcoded. This PR addresses the limitation by adding an optional field to the constructor to override the driver.

I've added a `postgresql-pg8000` extra preset for convenience, but I'm happy to revert if you find it redundant ( https://github.com/testcontainers/testcontainers-python/pull/156/commits/756f3e127836a12741bb629fb3c7e459b930bd94 ).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,6 +57,7 @@ setuptools.setup(
         'mysql': ['sqlalchemy', 'pymysql'],
         'oracle': ['sqlalchemy', 'cx_Oracle'],
         'postgresql': ['sqlalchemy', 'psycopg2-binary'],
+        'postgresql-pg8000': ['sqlalchemy', 'pg8000'],
         'selenium': ['selenium'],
         'google-cloud-pubsub': ['google-cloud-pubsub < 2'],
         'mongo': ['pymongo'],
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,6 @@ setuptools.setup(
         'mysql': ['sqlalchemy', 'pymysql'],
         'oracle': ['sqlalchemy', 'cx_Oracle'],
         'postgresql': ['sqlalchemy', 'psycopg2-binary'],
-        'postgresql-pg8000': ['sqlalchemy', 'pg8000'],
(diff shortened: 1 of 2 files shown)


KarolisL:
Btw, I wasn't able to quickly google the docstring format. Would you please point me to the spec so I could format the docstring properly? https://github.com/testcontainers/testcontainers-python/pull/156/files#diff-0d0163aaa012beed78914f91926197b1bd2fcd7fc84eaa6bf345367f2860d8cdR30

testcontainers/postgres.py
```diff
@@ -21,23 +21,29 @@ class PostgresContainer(DbContainer):
...
     -------
-    The example spins up a Postgres database and connects to it using the :code:`psycopg` driver.
+    The example spins up a Postgres database and connects to it using the (default) :code:`psycopg` driver.
     ::
 
         with PostgresContainer("postgres:9.5") as postgres:
             e = sqlalchemy.create_engine(postgres.get_connection_url())
             result = e.execute("select version()")
+    
+    ::
```
yakimka, line 31: I think that it is not necessary to change the example in the docstring.

yakimka (changes requested):

setup.py
```diff
@@ -57,6 +57,7 @@
         'mysql': ['sqlalchemy', 'pymysql'],
         'oracle': ['sqlalchemy', 'cx_Oracle'],
         'postgresql': ['sqlalchemy', 'psycopg2-binary'],
+        'postgresql-pg8000': ['sqlalchemy', 'pg8000'],
```
tillahoffmann, line 60: ```suggestion```

Users can install different drivers independently.

yakimka:
@KarolisL, please fix formatting

yakimka (approved):

PR #158, opened 2021-10-24
Improve host detection

tuvshuud:
This PR will fix additional issue that are reported in #43

Improvement:
- Improve host detection with docker internal hostname `host.docker.internal`
- Improve overall host detection in different scenarios

Bug fix:
- TC_HOST environment variable is not respected when set in docker in docker environment


yakimka:
@tuvshuud hello. Is this PR still valid? Do you have plans to continue working on this PR?

tuvshuud:
@yakimka Hello, I am willing to continue work on this, but not in near few months at least, I afraid.

totallyzen:
Hey folks, I know this might be an outstanding issue, but a lot of things have happened since this issue was even worked on.

I think it's not worth keeping this issue just on the basis that a few use-cases out there are broken.
I'm happy to open similar issues in the future, but we need actual live use-cases with reproducible examples in order to prioritise this, otherwise it's not valuable work.

Closing until further notice.

Issue #161, opened 2021-12-02
google cloud pubsub container do not expects keyword argument "channel" for publisher and subscriber client


Mohsin-Ul-Islam:
Whenever I create an instance of PubsubContainer from google.cloud module, it raises an exception "Unexpected keyword argument 'channel' for PublisherClient.__init__".

Mohsin-Ul-Islam:
This issue came when I used google cloud pubsub client library version 2.5.0 via poetry. While the requirements in this project contain version 1.7.0.

Here is the release where **channel** keyword argument was passed to underlying pubsub clients https://googleapis.dev/python/pubsub/0.43.0/_modules/google/cloud/pubsub_v1/subscriber/client.html
But this is outdated and the implementation has been changed.

Instead now **client_options** and **credentials** are passed to the underlying pubsub clients. Here is the version reference I used
https://googleapis.dev/python/pubsub/2.5.0/_modules/google/cloud/pubsub_v1/publisher/client.html

In order to bump google-cloud-pubsub from 1.7.0 to latest I get the following error while running
`pip-compile --verbose --upgrade-package google-cloud-pubsub==2.5.0`

Could not find a version that matches docker[ssh]<4.3.0,>=5 (from -r requirements.in (line 7))
[...]

PR #164, opened 2021-12-13
fix: pass updated keyword args to Publisher/Subscriber client in google/pubsub #161

Mohsin-Ul-Islam:
…le/pubsub


alexanderankin:
Seems like another fix could be a #446 release

PR #168, opened 2022-01-07
Build option for DockerCompose

aloosley:
There are some usecases where it is useful to force building to occur when using docker-compose (the equivalent to `docker-compose up --build`).  

In this PR, a boolean parameter is added to allow for this, similar to the `pull` boolean parameter that already exists to force image pulling.


yakimka:
@aloosley, hello. Can you update this PR?

testcontainers/compose.py
```diff
@@ -88,7 +90,7 @@ def start(self):
             pull_cmd = self.docker_compose_command() + ['pull']
             subprocess.call(pull_cmd, cwd=self.filepath)
 
-        up_cmd = self.docker_compose_command() + ['up', '-d']
+        up_cmd = self.docker_compose_command() + ['up', '-d'] + ["--build"] if self.build else []
```
tillahoffmann, line 93: I think the tests may be failing because of the operator precedence here (the ternary has higher precedence than the addition). Let's change this to:

```suggestion
        up_cmd = self.docker_compose_command() + ['up', '-d']
        if self.build:
            up_cmd.append('--build')
```

aloosley:
Hi both, thanks for checking in on this PR!  First glance, the needed change looks small.  I'll do this (and try to write a better test) in the next day when I get to a computer.

testcontainers/compose.py
aloosley: Fixed.

aloosley:
Alright @yakimka, I cleaned up the PR a bit.  Testing `build=True` would have been a bit complicated, so rather than add something complicated to this PR, I opted for the simple test option of simply making sure the arg `--build` actually gets used when `docker-compose` is called.  I hope this solution is ok.

tests/test_docker_compose.py
```diff
@@ -23,6 +25,22 @@ def test_can_pull_images_before_spawning_service_via_compose():
...
+    ) as call_mock:
+        with DockerCompose("tests", build=True) as compose:
+            host = compose.get_service_host("hub", 4444)
+            port = compose.get_service_port("hub", 4444)
+            assert host == "0.0.0.0"
+            assert port == "4444"
+
+    assert compose.build
+    docker_compose_cmd = call_mock.call_args_list[0][0][0]
+    assert "docker-compose" in docker_compose_cmd and "--build" in docker_compose_cmd
```
yakimka, line 41: I think this test can be simplified
```python
def test_can_build_images_before_spawning_service_via_compose():
    with patch.object(DockerCompose, "_run_command") as call_mock:
        with DockerCompose("tests", build=True):
            assert "--build" in call_mock.call_args_list[0].args[0]
```

It only needs to move the `subprocess.call` to a `DockerCompose` method and replace all `subprocess` calls in class:
```python
class DockerCompose(object):
...
    def _run_command(self, cmd):
        subprocess.call(cmd, cwd=self.filepath)
```

yakimka (changes requested):

testcontainers/compose.py
```diff
@@ -88,7 +90,7 @@ def start(self):
             pull_cmd = self.docker_compose_command() + ['pull']
             subprocess.call(pull_cmd, cwd=self.filepath)
 
-        up_cmd = self.docker_compose_command() + ['up', '-d']
+        up_cmd = self.docker_compose_command() + ['up', '-d'] + (["--build"] if self.build else [])
```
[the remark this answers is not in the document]
aloosley: Changed.

tests/test_docker_compose.py
aloosley: This test is weaker as it does not check that `build=True` still runs docker-compose up.  I at least add a check `docker-compose` and `up` are called in the docker-compose subprocess.  Look ok?

tests/test_docker_compose.py
```diff
@@ -23,6 +25,18 @@ def test_can_pull_images_before_spawning_service_via_compose():
         assert port == "4444"
 
 
+def test_can_build_images_before_spawning_service_via_compose():
+    with patch.object(DockerCompose, DockerCompose._call_command.__name__) as call_mock:
```
yakimka, line 29: No need to use `DockerCompose._call_command.__name__`, just use `"_call_command"`

tests/test_docker_compose.py
```diff
@@ -23,6 +25,18 @@ def test_can_pull_images_before_spawning_service_via_compose():
...
+        with DockerCompose("tests", build=True) as compose:
+            ...
+
+    assert compose.build
```
yakimka, line 33: This is an implementation detail. There is no need to test for this attribute because passing this attribute is tested below

tests/test_docker_compose.py
yakimka: Thank you. Just two small questions left (see below)

tests/test_docker_compose.py
aloosley: There is good reason to use `DockerCompose._call_command.__name__` vs. `"_call_command"` and that, when using a typical IDE to refactor the method name, most IDEs will update the test automatically using the former code, and cause the test to break using the suggestion.  

But I don't feel that strongly about it so I'll change it now.

tests/test_docker_compose.py
aloosley: It depends on if you see the attribute as a feature for users (i.e. you want users to have `compose.build == True` if it's passed at construction).  

But again, I don't feel strongly about it so I'll delete it.

yakimka: Hm, you convinced me. Let's leave it like this ;)

yakimka (approved):

aloosley:
🥳  Thanks @yakimka, any other steps needed here?

yakimka:
@aloosley thank you =)

Issue #140, opened 2021-04-29
Allow to specify dialect/driver when calling get_connection_url


inikolaev:
With latest SQLAlchemy 1.4 rolling out support for asyncio it would be great to support that in some way.

In order to use asyncio one needs to use a corresponding driver which is specified in the connection string, e.g.:

```
postgresql+asyncpg://scott:tiger@localhost/test
```

But in `PostgresContainer` it's currently hardcoded to:

```
postgresql+psycopg2
```

Don't know what the best solution could be:

* Pass it as a constructor argument - but in this case there would be a limitation that `get_connection_url` will return fixed connection string, which this might not be a huge issue realistically, but feels odd, since container doesn't really care how we want to connect to it
* Pass an argument to `get_connection_url`, but this messes up with an API provided by `DbContainer` and then other implementation should probably support this as well or at least ignore unknown parameters.

orsolli:
This is solved

tillahoffmann:
#156 for reference.

PR #175, opened 2022-01-15
Add support for asyncio

NEONKID:
This PR contains the core code prepared for asynchronous processing in Python.

Asynchronous processing using the asyncio module is supported from SQLAlchemy version 1.4.
(https://docs.sqlalchemy.org/en/14/orm/extensions/asyncio.html)

So, in order to use these asynchronous modules in testcontainers, we make a PR request for the code that supports these modules.

In addition to PostgreSQL, other relational DB drivers can also support it, and a parameter has been added to the function to select Dialect in the existing code.
(Python uses several DB driver, so I don't think it's good to hardcode a specific driver #140)

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -2,6 +2,7 @@
 codecov>=2.1.0
 flake8
 pytest
+pytest-asyncio
 pytest-cov
 sphinx
 docker<4.3.0
--- a/requirements/3.8.txt
+++ b/requirements/3.8.txt
@@ -108,6 +108,7 @@ protobuf==3.13.0
     #   google-api-core
     #   googleapis-common-protos
 psycopg2-binary==2.8.5
+asyncpg==0.25.0
     # via testcontainers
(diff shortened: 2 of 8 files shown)


NEONKID:
If you would like to provide feedback on this PR, feel free to leave a comment.

tillahoffmann:
Do you think it's feasible to achieve this using metaprogramming to avoid repeated code (one implementation for async one for sync)?

yakimka:
Are we need async implementation for testing purposes?

tillahoffmann:
I'm going to close this PR for now. The main reason is that the benefits of async only manifest in the async implementation of sqlalchemy. In particular, none of the actual container-related code uses `await` syntax such that the code will run exactly the same way as if it was entirely synchronous. Users who want to make use async database connections should instantiate their own async connection.

Feel free to disagree, and we can discuss reopening this PR.

PR #181, opened 2022-02-20
Set lower bound on docker version

naomielst:
From my experience, the docker version must be >= 4.0.0


thedrow (approved):

tillahoffmann:
This will also need a rebuild of the requirements (cf. #188). The motivation behind completely pinning down the requirements for tests is to be able to separate "issues caused by dependency versions changing" from "issues introduced in a PR".

tillahoffmann:
> I suggest using something like poetry instead and caching the lock file.

Yup, those requirements are locked down dependencies. We just generate them using [pip-tools](https://github.com/jazzband/pip-tools) rather than poetry or pipenv.

thedrow:
@tillahoffmann Would you mind doing it this time?

tillahoffmann:
See #203.

PR #183, opened 2022-03-13
Make the GCP pubsub container work

hy9be:
Changes

1. Replace ``with_exposed_ports`` with ``with_bind_ports``
2. Modify the context manager to set and unset environment variable as mentioned in https://cloud.google.com/pubsub/docs/emulator#env
3. Update the test

### DIFF
--- a/testcontainers/google/pubsub.py
+++ b/testcontainers/google/pubsub.py
@@ -13,6 +13,11 @@
 
 from ..core.container import DockerContainer
 
+from google.cloud import pubsub
+from google.auth import credentials
+
+import os
+
 
 class PubSubContainer(DockerContainer):
     """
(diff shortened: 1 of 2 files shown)


hy9be:
@SergeyPirogov @tillahoffmann could you please let me know how I can trigger the CI build to unblock the review? The status of 3.8 build is pending at ``Expected — Waiting for status to be reported``. Thanks.

hy9be:
@yakimka I updated my branch to fix the flake8 error in CI. Could you please help to trigger the CI again? Thanks.

hy9be:
The CI failed because all the ``google-cloud-pubsub`` is still on 1.x while the versions on PyPI is 2.x (https://pypi.org/project/google-cloud-pubsub/). Should I just update that in all the requirements file?

yakimka:
> The CI failed because all the `google-cloud-pubsub` is still on 1.x while the versions on PyPI is 2.x (https://pypi.org/project/google-cloud-pubsub/). Should I just update that in all the requirements file?

I think that @tillahoffmann may know why the version of google-cloud-pubsub is pinned

tillahoffmann:
There are changes that are not backwards compatible in the transition from 1.x to 2.x of the pubsub client, and we have bounded the pubsub client version above here. https://github.com/testcontainers/testcontainers-python/blob/9100aaf8701098db11792fad51116a9c8bf0cfd5/setup.py#L61

If your updated code works with 2.x, we can integrate the changes as follows:

- Update the requirement to be lower bounded, i.e. `>=2`.
- Run `make requirements` to update all the requirements files from the root directory.

hy9be:
@tillahoffmann thanks for the instructions. I updated the files accordingly. Local tests passed.

hy9be:
@tillahoffmann Thanks. I opened a new PR here https://github.com/testcontainers/testcontainers-python/pull/209. Please review. (Accidentally closed this one when resolving the conflicts)

PR #191, opened 2022-03-27
Fix flacky tests and improvements

yakimka:


testcontainers/neo4j.py
```diff
@@ -71,10 +70,13 @@ def get_connection_url(self):
...
     def _connect(self):
         deadline = time.time() + Neo4jContainer.NEO4J_STARTUP_TIMEOUT_SECONDS
-        regex = re.compile("Remote interface available at", re.MULTILINE).search
 
         # First we wait for Neo4j to say it's listening
-        wait_for_logs(self, regex, Neo4jContainer.NEO4J_STARTUP_TIMEOUT_SECONDS)
+        wait_for_logs(
+            self,
+            "Remote interface available at",
+            Neo4jContainer.NEO4J_STARTUP_TIMEOUT_SECONDS / 4,
```
tillahoffmann, line 78: What's the division here for?

yakimka: If we wait for `NEO4J_STARTUP_TIMEOUT_SECONDS`, then there is a moment when we wait 10 seconds for the log, but then the code falls because the deadline has come.

So the timeout for wait_for_logs must be less than `NEO4J_STARTUP_TIMEOUT_SECONDS

tillahoffmann: Yeah, this is a bit of a weird implementation because it first waits for logs and then checks for connections in a loop. Thoughts on this?

```python
# First we wait for Neo4j to say it's listening.
wait_for_logs(self, "Remote interface available at", Neo4jContainer.NEO4J_STARTUP_TIMEOUT_SECONDS)

# Then we actually check that the container really is listening.
with self.get_driver() as driver:
    # Drivers may or may not be lazy force them to do a round trip to confirm neo4j is working.
    with driver.session() as session:
        session.run("RETURN 1").single()
        return

raise TimeoutException(
    "Neo4j did not start within %.3f seconds" % Neo4jContainer.NEO4J_STARTUP_TIMEOUT_SECONDS
)
```

yakimka: Nice, but `raise TimeoutException` is unreachable. As for the rest - I like it.

PR #199, opened 2022-04-04
Pipeline improvements

### LINKED ISSUE #197: Readthedocs cant generate documentation for some modules
Readthedocs can't generate documentation for Clickhouse, Neo4j, and other modules that import non-required 3rd party libraries.

.readthedocs.yml 
```
python:
  version: 3.7
  install:
    - method: pip
      path: .
```

In Github actions make docs don't fail because we install all dependencies manually but in readthedocs environment error raised:

```
Warning, treated as error:
autodoc: failed to import class 'clickhouse.ClickHouseContainer' from module 'testcontainers'; the following exception was raised:
No module named 'clickhouse_driver'
```

We can fix this in two ways:
1. Import drivers always only in the method `_configure`
2. Specify all dependencies in the `.readthedocs.yml` file

Also, we can improve `Run checks` action

yakimka:
Fixes #197

Add all requirements for readthedocs (not sure if this will work).
Run linter, docs and codecov only once per pipeline


thedrow:
I strongly recommend using codecov for all versions of Python with branch coverage enabled.
From my experience with supporting multiple Python versions in Celery, that's how you detect you missed coverage for a specific workaround/backport in that Python version. Codecov knows how to handle that.

yakimka:
hm, thanks @thedrow, TIL 😄

tillahoffmann:
Looks good, just needs a rebase, I think.

Issue #104, opened 2020-07-01
Supporting Ryuk container


fsonntag:
So I'm using testcontainers within Java, as well as Python. I noticed that sometimes containers stick around in Python, while in Java they always get cleaned up properly thanks to [Ryuk](https://github.com/testcontainers/moby-ryuk).

I was wondering, if that was a decision made on purpose to not use Ryuk in the Python project? Otherwise supporting it would be really great, in order to keep the docker environments of the host machine clean.

tillahoffmann:
As far as I'm aware, there is no reason why we can't use ryuk. Contributions welcome!

fsonntag:
Thanks for the feedback, I will look into it when I can find some time. But ofc also wanted to make sure beforehand if there was a specific reason :)

tillahoffmann:
Closing this for now as #208 addresses a similar issue.

PR #208, opened 2022-04-24
Add ResourceCleaner for stop containers after exit (#104)

mrk-andreev:
I suggest to stop all containers after application stop. I used `atexit` like `JVMHookResourceReaper` in `testcontainers-java`.


testcontainers/core/docker_client.py
```diff
@@ -30,15 +32,17 @@ def run(self, image: str,
...
-        return self.client.containers.run(image,
-                                          command=command,
-                                          stdout=stdout,
-                                          stderr=stderr,
-                                          remove=remove,
-                                          detach=detach,
-                                          environment=environment,
-                                          ports=ports,
-                                          **kwargs)
+        return ResourceCleaner.instance().attach(
```
tillahoffmann, line 35: Thoughts on just registering the hook directly here rather than going via the `ResourceCleaner`? E.g. `atexit.register(container.stop)` (with some additional error checks).

mrk-andreev: Yes, thanks for your advice. Now code look more simpler. 

I combined `stop_silent` and `atexit.register(stop_silent(container))` because some times container does not exists in the end so we have to suppress `docker.errors.NotFound`. 

```python
def stop_silent(container):
    def wrapper():
        try:
            container.stop()
        except NotFound:
            pass
        except Exception as e:
            logger.exception(e)
    return wrapper
```

tillahoffmann: Nice! `atexit.register` even lets you pass in arguments. So you could further simplify to

```python
def stop_silent(container):
    try:
        container.stop()
    except NotFound:
        pass
    except Exception as e:
        logger.exception(e)
```

and then follow up with

```python
atexit.register(stop_silent, container)
```

mrk-andreev: Thanks, I applied your fix.

testcontainers/core/docker_client.py
```diff
@@ -30,15 +33,18 @@ def run(self, image: str,
...
-                                          **kwargs)
+        container = self.client.containers.run(image,
+                                       command=command,
+                                       stdout=stdout,
+                                       stderr=stderr,
+                                       remove=remove,
+                                       detach=detach,
+                                       environment=environment,
+                                       ports=ports,
+                                       **kwargs)
```
thedrow, line 44: Please fix the formatting.

testcontainers/core/cleaner.py
```diff
@@ -0,0 +1,14 @@
...
+logger = setup_logger(__name__)
+
+
+def stop_silent(container):
+    try:
+        container.stop()
+    except NotFound:
+        pass
+    except Exception as e:
+        logger.exception(e)
```
thedrow, line 14: Please add debug logs here so we'll have visibility on which containers are being shut down because they weren't already.

Also, does this work well with pytest-xdist?

thedrow (changes requested):

gaby:
@mrk-andreev Any updates on this?

PR #209, opened 2022-04-24
Make the GCP PubSub container work

hy9be:
Accidentally closed the PR under review https://github.com/testcontainers/testcontainers-python/pull/183 and cannot reopen it. Recreated this one.

Changes

- Replace with_exposed_ports with with_bind_ports
- Modify the context manager to set and unset environment variable as mentioned in https://cloud.google.com/pubsub/docs/emulator#env
- Update the test
- Update the pubsub gcp lib versions


tillahoffmann:
It looks like this PR is reverting a number of changes introduced in previous work, e.g the docker version from #203. This may be the result of copying code from an outdated branch.

hy9be:
Yeah, I think you are right. I will take a look.

PR #210, opened 2022-04-28
Allow Docker Client Configuration

### LINKED ISSUE #180: Expose docker.from_env() parameters
`testcontainers.core.DockerClient` calls `docker.from_env()` it it's constructor. This method has a few parameters that will be useful to be exposed (see [reference](https://docker-py.readthedocs.io/en/stable/client.html?highlight=from_env#docker.client.from_env))

Ideally `testcontainers.core.DockerClient` and conversely `testcontainers.core.DockerContainer` should expose these parameters.

timbmg:
Allows to control parameters passed to `DockerClient.from_env` to configure the docker client. Allowed parameters can be found in the [docker-py docs](https://docker-py.readthedocs.io/en/stable/client.html#docker.client.from_env).

Fixes #180


tillahoffmann:
Nice, are there sensible tests we can add to make sure this works as expected?

timbmg:
> Nice, are there sensible tests we can add to make sure this works as expected?

I have added tests that make sure that the `.from_env` is called properly. But, these are very high level. The functionality is implemented in the `docker` package itself, so not quite sure what better tests we could do. Any suggestions welcome though:)

tillahoffmann (commented):
Great, just one small suggestion.

PR #211, opened 2022-05-03
Add support for pg8000 postgres driver

orsolli:
Fix too specific transient exception

`OperationalError` inherits from `DBAPIError`.
`pg8000` driver raises `InterfaceError` which also inherits form `DBAPIError`

```
sqlalchemy.exc.InterfaceError: (pg8000.exceptions.InterfaceError) network error on read
(Background on this error at: https://sqlalche.me/e/14/rvf5)
```

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -1,5 +1,6 @@
 -e file:.[docker-compose,mysql,oracle,postgresql,selenium,google-cloud-pubsub,mongo,redis,mssqlserver,neo4j,kafka,rabbitmq,clickhouse]
 codecov>=2.1.0
+cryptography<37
 flake8<3.8.0  # 3.8.0 adds a dependency on importlib-metadata which conflicts with other packages.
 pg8000
 pytest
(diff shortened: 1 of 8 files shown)


orsolli:
I originally named it "Fix " because I already used the `pg8000` driver with the old version `3.4.2`, but with a custom `PostgresContainer` class inheriting from the original class.

orsolli:
If you won't merge this, I can use this overload instead:
```python
from sqlalchemy.exc import DBAPIError
from testcontainers.core.waiting_utils import wait_container_is_ready


class PostgresContainer(PostgresContainer):
    @wait_container_is_ready(*[DBAPIError])
    def _connect(self):
        import sqlalchemy
        engine = sqlalchemy.create_engine(self.get_connection_url())
        engine.connect()
```

tillahoffmann:
Changing to `DBAPEError` sounds like a great idea. Having said that, let's keep `pg8000` as a user-installed library rather than a default install (cf. #156).

orsolli:
> Changing to `DBAPEError` sounds like a great idea. Having said that, let's keep `pg8000` as a user-installed library rather than a default install (cf. #156).

Do you want me to cherry-pick 9efde6bbf77a5f03130eab0bdb46fbe7bd3b28d5 then and not have a test for it? Or is there a way to have pg8000 as a dev-requirement?

tillahoffmann:
You could add the `pg8000` requirement [here](https://github.com/testcontainers/testcontainers-python/blob/master/requirements.in) which will include it as a dev requirement for testing.

tillahoffmann (approved):

PR #212, opened 2022-05-05
Separate tests into different components for easier debugging.

tillahoffmann:
This PR sets up different test runs for different components to simplify inspection of test failures. Only the `core` tests will be required to pass for pull requests. Ideally, all tests will pass, but, in light of a variety of containers being added to the project, maintaining all of them has become a substantial commitment. The change should also speed up the pipeline because tests are run in parallel. It also adds more recent python versions to the test matrix.


yakimka (approved):
Nice work!

Issue #219, opened 2022-05-29
[oracle]: waiting for the listener...


pmpetit:
Hello,

In an OracleDB, you connect to the db through the oracle listener process. In this image, when oracle db starts, it also starts the listener that will establish cx between the client and the db.

in the file core/generic.py, when the container starts, it also try to connect in the same time (just after)
But in an oracle db, you have to wait for the listener.

```python
    def start(self):
        self._configure()
        super().start()
        self._connect()
        return self
```

it is very simple to test : set a breakpoint on the ._connect() and wait 5 sec before going to this step. You 'll see that it is ok.

if you don't, then it will fail, telling that 'service name can not be found or service name does not exists' but it is normal, listener is not running, it is starting...

i think maybe, we should remove this _connect(), or add a waiting loop before, but i'm not sure of the impact on the others rdbms.

pmpetit:
self = <sqlalchemy.dialects.oracle.cx_oracle.OracleDialect_cx_oracle object at 0x10aa4cd00>
cargs = ()
cparams = {'dsn': '(DESCRIPTION=(ADDRESS=(PROTOCOL=TCP)(HOST=localhost)(PORT=49154))(CONNECT_DATA=(SID=XE)))', 'password': 'oracle', 'user': 'system'}

    def connect(self, *cargs, **cparams):
        # inherits the docstring from interfaces.Dialect.connect
>       return self.dbapi.connect(*cargs, **cparams)
E       cx_Oracle.DatabaseError: ORA-12541: TNS:no listener

venv/lib/python3.9/site-packages/sqlalchemy/engine/default.py:597: DatabaseError

The above exception was the direct cause of the following exception:

pmpetit:
the error is here:
```python
    @wait_container_is_ready(*ADDITIONAL_TRANSIENT_ERRORS)
    def _connect(self):
        import sqlalchemy
        engine = sqlalchemy.create_engine(self.get_connection_url())
        engine.connect()
```

pmpetit:
here the log of the oracle container:
> Executing task in folder scribedb: docker logs --tail 1000 -f 656159bae802405480684738f905098291839a5e2bbf06d6e2d7e2a4efac420e <

Starting Oracle Net Listener.
Starting Oracle Database 11g Express Edition instance.


System altered.


we should wait for "System altered." to be sure the listener is running.

pmpetit:
No more issues with the last version.

PR #220, opened 2022-06-04
override _connect method for oracle only

pmpetit:
linked to #219


pmpetit:
it can be tested with (replace <path_to_instantclient_19_8> to match yours)
```python
    def test(self):
        cx_Oracle.init_oracle_client(lib_dir=f"<path_to_instantclient_19_8>")
        with OracleDbContainer() as oracle:
            url = oracle.get_connection_url()
            e = sqlalchemy.create_engine(url)
            result = e.execute("select 1 from dual")
            for row in result:
                self.assertEqual(row[0],1)
```

pmpetit:
in fact with the last verion, everything works. No need to merge.

PR #221, opened 2022-06-05
Add ArangoDB Container support

nshine:
ArangoDB -> https://www.arangodb.com

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -67,7 +67,7 @@ setuptools.setup(
         'rabbitmq': ['pika'],
         'clickhouse': ['clickhouse-driver'],
         'keycloak': ['python-keycloak'],
-        'arangodb': ['python-arango']
+        'arangodb': ['python-arango'],
     },
     long_description_content_type="text/x-rst",
     long_description=long_description,
--- a/setup.py
+++ b/setup.py
@@ -67,7 +67,7 @@ setuptools.setup(
         'rabbitmq': ['pika'],
         'clickhouse': ['clickhouse-driver'],
         'keycloak': ['python-keycloak'],
-        'arango':['python-arango']
+        'arangodb':['python-arango']
     },
     long_description_content_type="text/x-rst",
     long_description=long_description,
--- a/setup.py
+++ b/setup.py
@@ -67,7 +67,7 @@ setuptools.setup(
         'rabbitmq': ['pika'],
         'clickhouse': ['clickhouse-driver'],
         'keycloak': ['python-keycloak'],
-        'arangodb':['python-arango']
+        'arangodb': ['python-arango']
     },
     long_description_content_type="text/x-rst",
     long_description=long_description,
--- a/setup.py
+++ b/setup.py
@@ -67,6 +67,7 @@ setuptools.setup(
         'rabbitmq': ['pika'],
         'clickhouse': ['clickhouse-driver'],
         'keycloak': ['python-keycloak'],
+        'arango':['python-arango']
     },
     long_description_content_type="text/x-rst",
     long_description=long_description,
--- a/testcontainers/arangodb.py
+++ b/testcontainers/arangodb.py
@@ -28,7 +28,7 @@ class ArangoDbContainer(DbContainer):
     """
     def __init__(self, image="arangodb:latest", port_to_expose=8529, **kwargs):
         super().__init__(image=image)
-        self.port_to_expose = 8529
+        self.port_to_expose = port_to_expose
         self.with_exposed_ports(self.port_to_expose)
 
         self.arango_host = kwargs.get('ARANGO_HOST', 'localhost')
--- a/testcontainers/arangodb.py
+++ b/testcontainers/arangodb.py
@@ -1,0 +1,66 @@
+"""
+ArangoDB container support.
+"""
+from os import environ
+from arango import ArangoClient
+
+from testcontainers.core.generic import DbContainer
+from testcontainers.core.waiting_utils import wait_for_logs
+
+class ArangoDbContainer(DbContainer):
+    """
+    ArangoDB container.
+
+    Example
+    -------
+    The example will spin up a ArangoDB container.
+    You may use the :code:`get_connection_url()` method which returns a arangoclient-compatible url
+    in format :code:`scheme://host:port`. As of now, only a single host supported (over HTTP).
+    ::
+        with ArangoContainer("arangodb:3.9.1") as arango:
+            client = ArangoClient(hosts=arango.get_connection_url())
+
+            # Connect
+            sys_db = arango_client.db(username='root', password='')
+
+            # Create a new database named "test".
+            sys_db.create_database("test")
+    """
+    def __init__(self, image="arangodb:latest", **kwargs):
+        super().__init__(image=image)
+        self.port_to_expose = 8529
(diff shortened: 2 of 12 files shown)


thedrow:
Thanks!

Celery definitely needs this one for its integration suite.

tillahoffmann:
The failing tests seem to be the consequence of a recent sphinx change (https://github.com/sphinx-doc/sphinx/issues/10474). You should be able to fix it by changing `language = None` to `language = 'en'` in `docs/conf.py`.

nshine:
Fixed

setup.py
```diff
@@ -67,6 +67,7 @@
         'rabbitmq': ['pika'],
         'clickhouse': ['clickhouse-driver'],
         'keycloak': ['python-keycloak'],
+        'arangodb': ['python-arango']
```
tillahoffmann, line 70:
```suggestion
        'arangodb': ['python-arango'],
```
Nit: this just makes the line changes a bit easier to read because the comma is already there when the next container is added.

testcontainers/arangodb.py
```diff
@@ -0,0 +1,64 @@
...
+
+class ArangoDbContainer(DbContainer):
+    """
+    ArangoDB container.
+
+    Example
+    -------
+    The example will spin up a ArangoDB container.
+    You may use the :code:`get_connection_url()` method which returns a arangoclient-compatible url
+    in format :code:`scheme://host:port`. As of now, only a single host supported (over HTTP).
```
tillahoffmann, line 17:
```suggestion
    in format :code:`scheme://host:port`. As of now, only a single host is supported (over HTTP).
```

testcontainers/arangodb.py
```diff
@@ -0,0 +1,64 @@
...
+        with ArangoContainer("arangodb:3.9.1") as arango:
+            client = ArangoClient(hosts=arango.get_connection_url())
+
+            # Connect
+            sys_db = arango_client.db(username='root', password='')
+
+            # Create a new database named "test".
+            sys_db.create_database("test")
+    """
+    def __init__(self, image="arangodb:latest", **kwargs):
```
tillahoffmann, line 29:
```suggestion
    def __init__(self, image="arangodb:latest", port_to_expose=8529, **kwargs):
```

testcontainers/arangodb.py
```diff
@@ -0,0 +1,64 @@
...
+        super().__init__(image=image)
+        self.port_to_expose = 8529
```
tillahoffmann, line 31:
```suggestion
        self.port_to_expose = port_to_expose
```

testcontainers/arangodb.py
tillahoffmann: This makes it easier to modify the exposed port if necessary (we should also change that on a bunch of other containers at some point).

testcontainers/arangodb.py
```diff
@@ -0,0 +1,64 @@
...
+        self.with_exposed_ports(self.port_to_expose)
+
+        self.arango_host = kwargs.get('ARANGO_HOST', 'localhost')
+
+        # https://www.arangodb.com/docs/stable/deployment-single-instance-manual-start.html
+        self.arango_no_auth = kwargs.get(
```
tillahoffmann, line 37: Shall we also move these to proper keyword arguments? That should make it a bit easier for code completion, for example. Something like this?

```python
# arango_no_auth=None in the __init__ signature.
self.arango_no_auth = arango_no_auth or os.environ.get("ARANGO_NO_AUTH")
```

testcontainers/arangodb.py
```diff
@@ -0,0 +1,64 @@
...
+        port = self.get_exposed_port(self.port_to_expose)
+        url = f"{scheme}://{self.arango_host}:{port}"
+
+        return url
+
+    def _connect(self):
+        wait_for_logs(
+            self,
+            predicate="is ready for business",
+            timeout=10)
```
tillahoffmann, line 64: Let's use `TC_MAX_TRIES` as the timeout. It's somewhat obscurely hidden, sorry.

https://github.com/testcontainers/testcontainers-python/blob/f9b7374e637b042d97be4891b54a496e0f9f0b96/testcontainers/core/config.py#L3

tillahoffmann (commented):
Looks great! A few small comments.

nshine:
Thanks for your review.  updated.

testcontainers/arangodb.py
tillahoffmann: Makes sense. 👍 We should probably update that in mysql at some point.

testcontainers/arangodb.py
```diff
@@ -0,0 +1,72 @@
...
+
+            # Create a new database named "test".
+            sys_db.create_database("test")
+    """
+    def __init__(self,
+                 image="arangodb:latest",
+                 port_to_expose=8529,
+                 hostname='localhost',
+                 arango_root_password='passwd',
+                 arango_no_auth=None,
```
tillahoffmann, line 35: For `arango_no_auth` and `arango_random_root_password`, would it be possible to turn them into booleans (defaulting to `False`) to make the interface "more pythonic" (I appreciate it's not a very well defined concept)? Then the user doesn't have to know that passing a string of `1` will have the desired effect, but they can rely on the python semantics of `True` or `False`. What do you think?

testcontainers/arangodb.py
```diff
@@ -0,0 +1,72 @@
...
+from os import environ
+from testcontainers.core.config import MAX_TRIES
+from testcontainers.core.generic import DbContainer
+from testcontainers.core.waiting_utils import wait_for_logs
+
+
+class ArangoDbContainer(DbContainer):
+    """
+    ArangoDB container.
+
```
tillahoffmann, line 13: If we could document the parameters of the constructor here, that'd be great!

tillahoffmann (commented):
Few more thoughts.

testcontainers/arangodb.py
nshine: I think that by definition of falsy values also includes None (and empty collections, etc.) - I'm fine with aligning to your request - However I think even what we did here (None) is tricky concept and should be default to actual string of "0". 
for the following reasons:
1. aligning to ArangoDB documentation 
    https://www.arangodb.com/docs/stable/deployment-single-instance-manual-start.html

2. since it env var. we can't do os.environ['KEY'] = False/None (is it do expect string values) - I do understand that it not the same as it eventually passing to Docker through -e , but we do need to remember it originated in env variable . because we wrap the Docker python-sdk, it might be OK (they take care the convert of the Falsy env params)

TLDR;
- Resolved. however, I do have some concerns. (above)
- Removed additional input (hostname) -> not relevant as of now and can be "calculated" from matched internal func

nshine:
Hi, all resolved.

nshine:
@tillahoffmann - I will appreciate if you could merge and generate a new release.

nshine:
Hi @thedrow / @SergeyPirogov  any chance you can step in and approve&release this one ?

SergeyPirogov (approved):

SergeyPirogov:
@nshine thank you

nshine:
Hi @SergeyPirogov 
Thanks for the approve - can you please trigger a new release? i.e. 3.6.0

PR #222, opened 2022-06-14
Fix None/True/False handling in Arango.

tillahoffmann:
Minor changes to let `None` default to environment variables and `True/False` to be used as is. Cf. #221.


nshine:
Hi @tillahoffmann  / @thedrow  / @SergeyPirogov 
Can you please trigger new release ?

tillahoffmann:
@nshine, that should already be done. Let me know if you have a problem installing 3.6.0.

Issue #228, opened 2022-07-11
Unknown setting 'discovery.zen.minimum_master_nodes' in elasticsearch testcontainer 8.x


lippertto:
I am trying to start the testcontainers with a new 8.x elasticsearch image with testcontainers 3.6.0.
However, the container quits immediately, and I cannot execute any tests.

I found out that the  `discovery.zen.*` settings are no longer supported and will cause the container to quit now. [ES 8 breaking changes](https://www.elastic.co/guide/en/elasticsearch/reference/master/migrating-8.0.html#breaking-changes-8.0)
```
java.lang.IllegalArgumentException: unknown setting [discovery.zen.minimum_master_nodes] please check that any required plugins are installed, or check the breaking changes documentation for removed settings
```

Also, the check if the container is up will not work properly. I assume this is because Elasticsearch will use https now by default. I did not find anything in the changelogs, but there is a comment in the Java testcontainers here [ElasticsearchContainer.java] "major version 8 is secure by default" (https://github.com/testcontainers/testcontainers-java/blob/84d3444600aaeeac6813505fe5bfbf9ffce760ef/modules/elasticsearch/src/main/java/org/testcontainers/elasticsearch/ElasticsearchContainer.java#L178)

Here is the error message that I get.
```
received plaintext http traffic on an https channel, closing connection Netty4HttpChannel{localAddress=/10.199.0.2:9200, remoteAddress=/10.199.0.1:47096
```

In order for the container to start, I need to change the environment variables as follows:

This does not work
```python
from testcontainers.elasticsearch import ElasticSearchContainer

es = ElasticSearchContainer(image=f"elasticsearch:8.3.2")

try:
    es.start()
finally:
    es.stop()
```

If I remove the environment variable, this code works
```python
from testcontainers.elasticsearch import ElasticSearchContainer

es = ElasticSearchContainer(image=f"elasticsearch:8.3.2") \
    .with_env("xpack.security.enabled", "false")

es.env.pop('discovery.zen.minimum_master_nodes')

try:
    es.start()
finally:
    es.stop()
```

I could prepare a Pull Request in which I check whether the version of the container is 8 or higher and set the environment variables accordingly. Does this sound like the correct approach?

tillahoffmann:
Do you know whether simply removing `discovery.zen.minimum_master_nodes` would break elasticsearch tests for earlier versions?

lippertto:
[...]

PR #232, opened 2022-08-11
Support Elasticsearch 8.x

lippertto:
This PR addresses issue #228

### DIFF
--- a/testcontainers/elasticsearch.py
+++ b/testcontainers/elasticsearch.py
@@ -10,10 +10,50 @@
 #    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 #    License for the specific language governing permissions and limitations
 #    under the License.
+import logging
+import re
+import urllib
+from typing import Dict
+
 from deprecation import deprecated
+
 from testcontainers.core.container import DockerContainer
 from testcontainers.core.waiting_utils import wait_container_is_ready
-import urllib
+
+_FALLBACK_VERSION = 8
+"""This version is used when no version could be detected from the image name."""
+
+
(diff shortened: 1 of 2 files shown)


testcontainers/elasticsearch.py
```diff
@@ -10,10 +10,50 @@
...
+        logging.warning("Could not determine major version from image name '%s'. Will use %s",
+                image_name, _FALLBACK_VERSION)
+        return _FALLBACK_VERSION
+    else:
+        return int(regex_match.group(1))
+
+
+def _environment_by_version(version: int) -> Dict[str, str]:
+    """Returns environment variables required for each major version to work."""
+    if version == 6:
```
tillahoffmann, line 44: Thoughts on using `version <= 6` here for backwards compatibility?

tillahoffmann (commented):
Looks good, one small question.

testcontainers/elasticsearch.py
lippertto: elastic.co supports only three major versions at a time, and version 5 has been eol for 2.5 years.

But I have no strong opinion here, and I understand that there will be people out there who still use version 5. (Although, I doubt that they will use testcontainers)

What I can do is to add a separate case for versions <= 5 . For these versions I would use the same setting as for version 6 and log a warning that these are neither supported by elastic.co nor this project. What do you think?

lippertto: I have tested version `5.6.16`, and that runs fine with the environment for version 6

tillahoffmann: Okay, let's leave it as-is.

tillahoffmann:
Thanks for fixing this!

PR #240, opened 2022-08-28
Make DockerCompose.get_service_info public

csikb:
I'm new to Python. I welcome every suggestion. :)

### DIFF
--- a/testcontainers/compose.py
+++ b/testcontainers/compose.py
@@ -227,7 +227,7 @@ class DockerCompose(object):
         result = str(output).rstrip().split(":")
         if len(result) != 2 or not all(result):
             raise NoSuchPortExposed(f"port {port} is not exposed for service {service}")
-        return result[0], result[1]
+        return result
 
     def _call_command(self, cmd, filepath=None):
         if filepath is None:
--- a/testcontainers/compose.py
+++ b/testcontainers/compose.py
@@ -186,7 +186,7 @@ class DockerCompose(object):
         str:
             The mapped port on the host
         """
-        return self._get_service_info(service_name, port)[1]
+        return self.get_service_info(service_name, port)[1]
 
     def get_service_host(self, service_name, port):
         """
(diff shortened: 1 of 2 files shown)


testcontainers/compose.py
```diff
@@ -204,15 +204,30 @@ def get_service_host(self, service_name, port):
...
+        (str, str):
+            The hostname for the service, The port for the service
+        """
         port_cmd = self.docker_compose_command() + ["port", service, str(port)]
         output = subprocess.check_output(port_cmd, cwd=self.filepath).decode("utf-8")
         result = str(output).rstrip().split(":")
         if len(result) != 2 or not all(result):
             raise NoSuchPortExposed(f"port {port} is not exposed for service {service}")
-        return result
+        return result[0], result[1]
```
tillahoffmann, line 230:
```suggestion
        return result
```

Shall we keep this as is, avoiding the extra indexing?

csikb: I only did this so the return type would be a `(str, str)` string touple and not a string array

csikb: I was playing with the idea to introduce a `HostPort = collections.namedtouple("HostPort", ["host", "port])` and wrap the result there

tillahoffmann: Let's keep this as is for now unless there's a particular reason to return a tuple? If a tuple is required, let's use `tuple(result)`.

csikb:
In the future it would make sense to depreciate the get_host and get_port methods in favour of this.

testcontainers/compose.py
csikb: @tillahoffmann I commited your suggestions. Can you please approve :)

tests/test_core/test_docker_compose.py
```diff
@@ -64,8 +64,7 @@ def test_can_parse_multiple_compose_files():
         assert host == "0.0.0.0"
         assert port == "3306"
 
-        host = compose.get_service_host("hub", 4444)
```
tillahoffmann, line 67: One more small change: let's check both methods for getting the information here to ensure we main backwards compatibility.

alexanderankin:
hi @csikb unfortunately this module has changed and your PR doesn't land so neatly - feel free to reply to this issue if this API is not sufficient as it seems to be similar in intent: `with DockerCompose() as d: d.get_container(service_name).get_publisher(by_port=port)` API.

it returns this object which contains the `URL` and `PublishedPort`

https://github.com/testcontainers/testcontainers-python/blob/5bef18a51360a2d74ba393f86b753abdf9ec5636/core/testcontainers/compose/compose.py#L29-L38

unfortunately it doesn't quite work as expected on #457 (we normalize the host in the public API of the module), arguably we should normalize it "sooner"?

PR #245, opened 2022-09-13
Configurable MongoDB exposed port

RubenRibGarcia:
Added port_to_expose argument on __init__ to configure MongoDB with different exposed ports


tillahoffmann (approved):
Great, thank you!

PR #250, opened 2022-09-29
Add MockServer container

Xeonian:
Add a MockServer container class.

Uses the `mockserver` client library to supply a client.

Had to pin `grpcio-status` dependency for the `google-cloud-pubsub` extra as the latest version produces conflicting requirements for the `protobuf` dependency.


testcontainers/mockserver.py
```diff
@@ -0,0 +1,50 @@
...
+            response = requests.get(f'{mockserver.base_url}/path')
+            assert response.status_code == HTTPStatus.OK
+
+            mockserver_client.verify()
+    """
+
+    EDGE_PORT = 1080
+    IMAGE = 'mockserver/mockserver:latest'
+
+    def __init__(self, image: str = IMAGE, **kwargs: Any):
```
pffijt, line 32: Would it be better to make use of the port argument so one can overwrite the port? 
Since the port can be changed with an environment variable: https://www.mock-server.com/where/docker.html#docker_container_environment_variables

testcontainers/mockserver.py
```diff
@@ -0,0 +1,50 @@
...
+
+class MockServerContainer(DockerContainer):
+    """
+    MockServer container.
+
+    Example
+    -------
+    ::
+
+        with MockServerContainer() as mockserver:
```
pffijt, line 18: I think it would be better to document what the example does?

Yourgene:
Hello!
What is the status of this PR, as I see it have no activities since 2023.

Recently I had to write for work a MockServer container class for work. If nobody is working on it anymore, I can maybe find some time to port my written code here

PR #252, opened 2022-09-30
Pass `**kwargs` from MySqlContainer constructor to DbContainer constructor

skirino:
Currently `MySqlContainer` does not pass `kwargs` to its base class, so we cannot customize docker client's timeout. This PR enables the users of `MySqlContainer` to set docker client's configurations via `docker_client_kw`.

Context:
We encountered read timeout errors during creation of mysql containers in our CI:
```
urllib3.exceptions.ReadTimeoutError: HTTPConnectionPool(host='localhost', port=2375): Read timed out. (read timeout=60)
```
We are still not sure what's the cause, but our current guess is that it's simply due to slow network.

### DIFF
--- a/testcontainers/mysql.py
+++ b/testcontainers/mysql.py
@@ -34,15 +34,20 @@ class MySqlContainer(DbContainer):
             version, = result.fetchone()
     """
 
-    def __init__(self, image="mysql:latest", **kwargs):
+    def __init__(self,
+                 image="mysql:latest",
+                 MYSQL_USER=None,
+                 MYSQL_ROOT_PASSWORD=None,
+                 MYSQL_PASSWORD=None,
+                 MYSQL_DATABASE=None,
+                 **kwargs):
         super(MySqlContainer, self).__init__(image, **kwargs)
         self.port_to_expose = 3306
         self.with_exposed_ports(self.port_to_expose)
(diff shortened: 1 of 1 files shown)


testcontainers/mysql.py
```diff
@@ -35,7 +35,7 @@ class MySqlContainer(DbContainer):
     """
 
     def __init__(self, image="mysql:latest", **kwargs):
-        super(MySqlContainer, self).__init__(image)
+        super(MySqlContainer, self).__init__(image, **kwargs)
```
tillahoffmann, line 38: It looks like, if `MYSQL_USER` or a similar variable is provided, these would also be passed up the inheritance chain.

tillahoffmann (commented):
Thanks for the contribution. I've left one small comment.

testcontainers/mysql.py
skirino: Changed the constructor parameter list so that `MYSQL_*` are passed as separate arguments and not included in `kwargs`: https://github.com/testcontainers/testcontainers-python/pull/252/commits/322e55f1cad1c181e593ecb03d8bc40be0fefe9e
Please let me know what you think!

tillahoffmann (approved):
Looks great!

PR #254, opened 2022-10-13
Update compose to work with both "docker-compose" as well as "docker compose"

jhnnsrs:
Fixed using docker compose with the new "docker compose" api as well as the older "docker-compose" api.

### DIFF
--- a/neo4j/testcontainers/neo4j/__init__.py
+++ b/neo4j/testcontainers/neo4j/__init__.py
@@ -76,8 +76,7 @@ class Neo4jContainer(DbContainer):
         with self.get_driver() as driver:
             # Drivers may or may not be lazy
             # force them to do a round trip to confirm neo4j is working
-            with driver.session() as session:
-                session.run("RETURN 1").single()
+            driver.verify_connectivity()
 
     def get_driver(self, **kwargs) -> Driver:
         return GraphDatabase.driver(


jooola:
Fixes #306

testcontainers/compose.py
```diff
@@ -89,6 +89,19 @@ def __enter__(self):
...
+        """
+        Returns the basecommand parts used for the docker compose commands depending on the docker compose api
+
+        Returns
+        -------
+        list[str]
+            The docker compose command parts
+        """
+        return ["docker","compose"] if subprocess.run(["docker", "compose", "--help"], stdout=subprocess.DEVNULL,
+        stderr=subprocess.STDOUT).returncode == 0 else ["docker-compose"]
```
jooola, line 104: I wonder if this should be configurable by the user ? There might be cases where the user prefers `docker-compose` over `docker compose`. Maybe the magic here is good, but as fallback from a user provided setting?

tillahoffmann: That sounds reasonable, maybe we can include a `compose_command` argument in the constructor. If it is `None`, we can use the magic to figure out the right one. Otherwise the provided value will be used as-is.

ash1425: @tillahoffmann Given that we have moved all containers to their directory, should we rebase this PR ? Maybe one of us will have to do it if OP is not around.

jhnnsrs: Accidentally closed this PR and opened #312 which includes the suggested changes

PR #258, opened 2022-10-24
raise an exception when docker compose fails to start

bitplane:
## What happened
If my docker-compose.yml has errors and I'm testing it using testcontainers, no error is reported

## What should happen
A `ContainerStartException` should be raised if docker-compose fails to start.

## Fix
Get the return code, raise exception if nonzero. See test and small change

### DIFF
--- a/compose/testcontainers/compose/__init__.py
+++ b/compose/testcontainers/compose/__init__.py
@@ -10,7 +10,7 @@ import subprocess
 from typing import Iterable, List, Optional, Tuple, Union
 
 from testcontainers.core.waiting_utils import wait_container_is_ready
-from testcontainers.core.exceptions import NoSuchPortExposed
+from testcontainers.core.exceptions import ContainerStartException, NoSuchPortExposed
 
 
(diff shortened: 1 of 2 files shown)


tillahoffmann (approved):
Great, thank you!

bitplane:
In case it's helpful to anyone else, to fix this old patch I did this:

1. Rebased to get rid of any changes after mine (I'd done `git merge master` a couple of times, so had to rebase on top of the change after the last merge commit). `THAT_COMMIT` is now the last change before mine were applied
2. Move the `git diff THAT_COMMIT --name-only` files into the new place with `git mv`
3. Commit the changes, then `git rebase -i THAT_COMMIT` again, and rearrange it so the commit with the `git mv` happens before all other commits
4. Now `git rebase master`. Get a merge conflict on the mv commit, so skip it

My subsequent patches applied cleanly with no merge conflicts. :)

tillahoffmann (approved):

alexanderankin:
closing in favor of #485

Issue #257, opened 2022-10-24
Clickhouse container ports setup


SomeAkk:
When we use Clickhouse, we can use 2 kind of ports: 9000 and 8123.
Also different awesome libs work 

- with just 9000
- with just 8123
- with both 8123 and 9000 ports

Problem in that: i can use 8123 port or 9000, but can't proper expose it (for example outside port must be 66612) because exposing port seted auto and than no overrided by `with_bind_ports()`. Also it can be needed to expose 1823 and 9000 together.

```
ch_container = ClickHouseContainer(image='clickhouse/clickhouse-server:22.6.4', port=8123)
ch_container.with_bind_ports(8123, 66612)
db = ch_container.start() <- started and expose 8123 port to random port 69112 (not 66612)
print(ch.get_connection_url()) <- gives url with port 69112 (not 66612)
```

tillahoffmann:
@yakimka, do you have thoughts on this?

yakimka:
@tillahoffmann I created two PRs, please have a look

SomeAkk:
@tillahoffmann @yakimka - any news for pr?

PR #261, opened 2022-10-31
Draft: Add ability to expose both http and native ports for clickhouse (#257)

yakimka:


tillahoffmann:
@yakimka, sorry for the slow follow up on this. Would you mind rebasing on `master`? I'm also planning a major-version release so we can squeeze in the breaking change of dropping the `port` argument. I'm breaking a bunch of other things in #296.

PR #262, opened 2022-11-05
Workaround for docker-in-docker

Duchadian:
## Previous behaviour

In situations where the `DOCKER_HOST` env variable is set, `DockerContainer.get_container_host_ip()` returned either the gateway ip or the bridge ip instead of the actual container host ip. This caused connection issues in a docker-in-docker context, see #217. 

## Changed behaviour

The `DockerContainer.get_container_host_ip()` function now uses the container ip whether the `DOCKER_HOST` variable is set in the environment. This should fix docker-in-docker situations while leaving other scenarios in a working state. Additionally, a test has been added that attempts to simulate a docker-in-docker situation. 

## Notes

This PR is not intended to be anything more than a workaround for the current docker-in-docker issue (#217). I've intentionally kept the size and scope of the PR small to avoid making this a PR to fix all networking-related issues. It's by no means perfect, and any feedback and input on both the workaround and the test case is appreciated.

### DIFF
--- a/testcontainers/core/container.py
+++ b/testcontainers/core/container.py
@@ -1,3 +1,4 @@
+import os
 from deprecation import deprecated
 from docker.models.containers import Container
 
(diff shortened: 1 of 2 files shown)


tillahoffmann:
Thank you, this seems like a good start to addressing the various networking issues we still have!

Issue #126, opened 2020-12-22
TimeScaleDb container


vikrambosevibrent:
TimeScaleDb Container not available. is that any plan to add that? Thanks

tillahoffmann:
Not planned at the moment, but PRs are welcome.

vikrambosevibrent:
sure.. Will do that..

pffijt:
It wouldn't be necessary to create a separate module/client for a TimescaleDB container since TimescaleDB is just an extension on top of PostgreSQL [(see here)](https://docs.timescale.com/getting-started/latest/#what-is-timescaledb).  Users of the TimescaleDB container can make use of PostgresDB container client.

Example:
```python
timescaled_container = PostgresContainer('timescale/timescaledb:latest-pg12')

with timescaled_container as timescaled:
    e = sqlalchemy.create_engine(timescaled.get_connection_url())
    e.execute('''
        CREATE TABLE stocks_real_time (
          time TIMESTAMPTZ NOT NULL,
          symbol TEXT NOT NULL,
          price DOUBLE PRECISION NULL,
          day_volume INT NULL
        ); 
    ''')

    e.execute('''
        SELECT create_hypertable('stocks_real_time','time');
    ''')
```

pffijt:
I think we can close this issue :)

PR #265, opened 2022-11-14
adding support for dragonfly

boazsade:
This would add support to test [dragonfly](https://dragonflydb.io/) using test-containers-python
This include:
- new entry in the test containers list.
- update the readme file.
- update github workflow.
Note that since dragonfly work out of the box with the same clients as Redis, there is no requirement to change the installations or dependencies here.
Signed-off-by: Boaz Sade <boaz@dragonflydb.io>

### DIFF
--- a/testcontainers/dragonfly.py
+++ b/testcontainers/dragonfly.py
@@ -1,0 +1,58 @@
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


testcontainers/dragonfly.py
```diff
@@ -0,0 +1,57 @@
...
+        ----------
+        kwargs: dict
+            Keyword arguments passed to `redis.Redis`.
+
+        Returns
+        -------
+        client: redis.Redis
+            Redis client to connect to the container.
+        """
+        return redis.Redis(
```
tillahoffmann, line 47: Could this be implemented by reusing the redis container and changing the image in the constructor?

boazsade: I think it would make it more clear that this is not redis - even though you're using the same client to connect to

tillahoffmann:
Closing this PR for now because we can use the redis container with a different image (see #126 for a similar discussion).

PR #280, opened 2022-12-15
Fix typo in environment variable and add warning about usage

mattupstate:
There's an obvious typo in the environment variable. This PR updates the name to reflect the behavior and adds a warning about using the old one.


tillahoffmann:
Hi @mattupstate, thanks for the fix. It looks like there are a few small linting errors. Are you up for rebasing on `master`?

Issue #108, opened 2020-07-05
could not translate host name "localnpipe" to address: Unknown host


dbaskarraj:
venv\lib\site-packages\wrapt\wrappers.py:605: in __call__
    return self._self_wrapper(self.__wrapped__, self._self_instance,
venv\lib\site-packages\testcontainers\core\waiting_utils.py:46: in wrapper
    raise TimeoutException(
E   testcontainers.core.exceptions.TimeoutException: Wait time exceeded 120 sec.
E                       Method _connect, args () , kwargs {}.
E                        Exception (psycopg2.OperationalError) could not translate host name "localnpipe" to address: Unknown host
E
E   (Background on this error at: http://sqlalche.me/e/e3q8)

Please help me out with this thing.

tillahoffmann:
@arpost, do you know whether this issue is related to the fix you added in #100?

petedmarsh:
I am experiencing the same issue on Windows 10. If I subclass `PostgresContainer` like so then it works:

```
class X(PostgresContainer):
    def get_connection_url(self):
        return super().get_connection_url().replace('localnpipe', 'localhost')

with X('postgres:12.3') as pg:
    print("HELLO")
    engine = sqlalchemy.create_engine(pg.get_connection_url())
    version, = engine.execute("select version()").fetchone()
    print(version) 
```

i.e. `postgresql+psycopg2://test:test@localhost:32779/test` does work as the URL.

Some quick Googling suggests there are issues with Windows 10 and Docker named pipes but I can't honestly say they are the ultimate cause of this issue.

Perhaps `DockerClient.host()` could be changed to return `localhost` if the host is Windows 10? Happy to submit a patch for that but again I am not 100% that is the "right" thing to do.

greg-el:
> 
> 
> I am experiencing the same issue on Windows 10. If I subclass `PostgresContainer` like so then it works:
> 
> ```
> class X(PostgresContainer):
>     def get_connection_url(self):
>         return super().get_connection_url().replace('localnpipe', 'localhost')
> 
> with X('postgres:12.3') as pg:
>     print("HELLO")
>     engine = sqlalchemy.create_engine(pg.get_connection_url())
>     version, = engine.execute("select version()").fetchone()
>     print(version) 
> ```
> 
> i.e. `postgresql+psycopg2://test:test@localhost:32779/test` does work as the URL.
> 
[...]

PR #283, opened 2022-12-26
fix DinD host detection when using local socket

ikravets:
When using IPC socket (unix/npipe) to communicate with local docker server docker-py implementation rewrites the base url to HTTP, see https://github.com/docker/docker-py/blob/e901eac7a8c5f29c7720eafb9f58c8356cca2324/docker/api/client.py#L143-L168

We cannot rely on URL scheme in order to detect such connections. Instead, we detect connection adapters added by docker-py api client:

- UnixHTTPAdapter https://github.com/docker/docker-py/blob/e901eac7a8c5f29c7720eafb9f58c8356cca2324/docker/transport/unixconn.py

- NpipeHTTPAdapter https://github.com/docker/docker-py/blob/e901eac7a8c5f29c7720eafb9f58c8356cca2324/docker/transport/npipeconn.py

As NpipeHTTPAdapter type may not be available outside of Windows we rely on adapter-specific attributes.

### DIFF
--- a/testcontainers/core/docker_client.py
+++ b/testcontainers/core/docker_client.py
@@ -14,6 +14,7 @@ import os
 import urllib
 import docker
 from docker.models.containers import Container
+from docker.transport import UnixHTTPAdapter
 from testcontainers.core.utils import inside_container
 from testcontainers.core.utils import default_gateway_ip
 
@@ -72,11 +73,14 @@ class DockerClient(object):
 
         except ValueError:
             return None
+        adapter = self.client.api.get_adapter(self.client.api.base_url)
+        is_ipc = isinstance(adapter, UnixHTTPAdapter)
+        is_ipc |= hasattr(adapter, "socket_path") or hasattr(adapter, "npipe_path")
+        is_ipc |= 'unix' in url.scheme or 'npipe' in url.scheme
+        if is_ipc and inside_container():
+            ip_address = default_gateway_ip()
+            if ip_address:
+                return ip_address
         if 'http' in url.scheme or 'tcp' in url.scheme:
             return url.hostname
-        if 'unix' in url.scheme or 'npipe' in url.scheme:
-            if inside_container():
-                ip_address = default_gateway_ip()
-                if ip_address:
-                    return ip_address
         return "localhost"


ikravets:
Rebased to the master. @tillahoffmann, any chance for a review?

tillahoffmann:
Thank you for these improvements! Are there tests that we could add that fail before the change is introduced to avoid future regressions?

core/testcontainers/core/docker_client.py
```diff
@@ -102,11 +103,14 @@ def host(self) -> str:
 
         except ValueError:
             return None
+        adapter = self.client.api.get_adapter(self.client.api.base_url)
+        is_ipc = isinstance(adapter, UnixHTTPAdapter)
+        is_ipc |= hasattr(adapter, "socket_path") or hasattr(adapter, "npipe_path")
+        is_ipc |= 'unix' in url.scheme or 'npipe' in url.scheme
```
tillahoffmann, line 109: Even if the `NpipeHTTPAdapter` is only used on windows, would it still be available for an `isinstance` check independent of the platform? Or does the `docker` package not even expose `NpipeHTTPAdapter` if it's not on windows?

Alternatively, would checking `adapter.__class__.name == 'NpipeHTTPAdapter'` be easier to understand than checking the attributes of the adapter?

tillahoffmann (commented):
This is great, thank you for having such a detailed look into improving the host detection. Host detection is the challenge that seems to cause most trouble for our users so far.

Do you know whether this change would also resolve the issue of `localnpipe` appearing in the connection string, as discussed in https://github.com/testcontainers/testcontainers-python/issues/108? You're much more clued up on this than I am.

core/testcontainers/core/docker_client.py
ikravets: I don't think that `NpipeHTTPAdapter` symbol is available outside of windows. Checking the class name would break for `NpipeHTTPAdapter` subclasses. So I went on with a classic duck-typing.

ikravets:
@tillahoffmann Hi, it's been more than half a year since I've opened this PR. , Can we maybe merge this?

ikravets:
@tillahoffmann a kind ping

CarliJoy:
This should be obsolete now that #714 is merged.
Could you check?
If you think the detection is still required maybe add a `is_socket_connection` method or property and include the results of the check to either `host` or `get_connection_mode` of the DockerClient.

core/testcontainers/core/docker_client.py
CarliJoy: I think something like this:

```python
try:
    from docker.transport.npipeconn import NpipeHTTPAdapter
except ImportError:
    class NpipeHTTPAdapter: ...
```

Would be a more proper solution.

It is much clearer what happens, does not depend the design of the Adapter and respects subclasses.

PR #284, opened 2022-12-26
make dev version PEP 440 compatible

ikravets:
Some package managers (e.g. poetry) may enforce PEP 440 even when installing development-only version. This change supports such a use case.


ikravets:
Rebased to the master. @tillahoffmann, any chance for a review?

PR #234, opened 2022-08-22
Add azurite module

pffijt:
Pull request to add azurite module. https://hub.docker.com/_/microsoft-azure-storage-azurite

PR #296, opened 2023-01-07
Remove class-level variables from containers.

tillahoffmann:
Several of the containers currently load environment variables at import time. Consequently, if we create a container, its configuration may not match the current environment variables. Containers also overwrite class variables with instance variables of the same name which can cause confusion. 

This PR 
- removes all class-level variables and loads environment variables at the time the container is created.
- unifies the container API by renaming `user` to `username` and `port_to_expose` to `port` where applicable such that all containers have the same parameter names. The motivation behind renaming `port_to_expose` is that some containers have multiple different ports (cf. #257 and #234). `[specific]_port_to_expose` is just a bit long.

This *will* break a bunch of stuff, but I'm planning to squeeze a lot of this cleanup into the next major release.

### DIFF
--- a/neo4j/testcontainers/neo4j/__init__.py
+++ b/neo4j/testcontainers/neo4j/__init__.py
@@ -40,7 +40,7 @@ class Neo4jContainer(DbContainer):
     def __init__(self, image: str = "neo4j:latest", *, bolt_port: int = 7687,
                  password: Optional[str] = None, username: Optional[str] = None, **kwargs) -> None:
         super(Neo4jContainer, self).__init__(image, **kwargs)
-        self.username = username or os.environ.get("NEO4J_USER", "password")
+        self.username = username or os.environ.get("NEO4J_USER", "neo4j")
         self.password = password or os.environ.get("NEO4J_PASSWORD", "password")
         self.bolt_port = bolt_port
         self.with_exposed_ports(self.bolt_port)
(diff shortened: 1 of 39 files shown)


neo4j/testcontainers/neo4j/__init__.py
```diff
@@ -35,42 +38,25 @@ class Neo4jContainer(DbContainer):
...
-    # The official image requires a change of password on startup.
-    NEO4J_ADMIN_PASSWORD = os.environ.get("NEO4J_ADMIN_PASSWORD", "password")
-    # Default port for the binary Bolt protocol.
-    DEFAULT_BOLT_PORT = 7687
-    AUTH_FORMAT = "neo4j/{password}"
-    NEO4J_STARTUP_TIMEOUT_SECONDS = 10
-    NEO4J_USER = "neo4j"
-
-    def __init__(self, image: str = "neo4j:latest", **kwargs) -> None:
+    def __init__(self, image: str = "neo4j:latest", *, port: int = 7687,
```
pffijt, line 41: We could the remove the asterisk argument to keep consistency.

pffijt (approved):

pffijt:
We should fix the minor inconsistencies then the pr would be ready for merge.

ImFlog (approved):

pffijt (approved):

PR #298, opened 2023-01-23
Allow to use dialect mssql+pyodbc

gplssm:


mssql/testcontainers/mssql/__init__.py
```diff
@@ -39,7 +54,21 @@ def _configure(self) -> None:
...
         )
+        if self.dialect == "mssql+pyodbc":
+            import pyodbc
+            import re
+            r = re.compile('ODBC Driver \d{1,2} for SQL Server')
+            # We sort drivers in reversed order to get the latest
+            drivers = sorted(list(filter(r.match, pyodbc.drivers())), reverse=True)
+
+            if len(drivers) > 0:
+                driver_str = drivers[0].replace(" ", "+")
```
tillahoffmann, line 69: My hunch is that this won't work because `sorted` is an iterable that can't be indexed. We likely also need to parse the version as an integer because, using string comparison, version 8 is greater than version 10. E.g.,

```python
>>> max(['ODBC Driver 10 for SQL Server', 'ODBC Driver 8 for SQL Server'])
'ODBC Driver 8 for SQL Server'
```

mssql/testcontainers/mssql/__init__.py
```diff
@@ -36,10 +58,32 @@ def _configure(self) -> None:
...
+            url += f"?driver={self._get_url_suffix_for_latest_pyodbc_version()}"
+        return url
+
+    def _get_url_suffix_for_latest_pyodbc_version(self):
+        import pyodbc
+        import re
+
+        r = re.compile(r"ODBC Driver \d{1,2} for SQL Server")
+        # We sort drivers in reversed order to get the latest
+        drivers = sorted(list(filter(r.match, pyodbc.drivers())), reverse=True)
```
tillahoffmann, line 81: We'll have to convert to `int` before sorting.

tillahoffmann:
The test fails because `Literal` is not available in python 3.7 (cf https://github.com/python/typing/issues/707). As 3.7 is close to end of life, let's omit the `Literal` typing for now, and we can re-introduce it once we drop support for python 3.7.

PR #300, opened 2023-01-27
Neo4j: verify_connectivity for connection test

robsdedude:
The driver offers an API specifically crafted to test if it can successfully connect. This tends to exchange less data than a dummy query and hence should be slightly faster.

See also https://neo4j.com/docs/api/python-driver/current/api.html#neo4j.Driver.verify_connectivity

### DIFF
--- a/neo4j/testcontainers/neo4j/__init__.py
+++ b/neo4j/testcontainers/neo4j/__init__.py
@@ -76,8 +76,7 @@ class Neo4jContainer(DbContainer):
         with self.get_driver() as driver:
             # Drivers may or may not be lazy
             # force them to do a round trip to confirm neo4j is working
-            with driver.session() as session:
-                session.run("RETURN 1").single()
+            driver.verify_connectivity()
 
     def get_driver(self, **kwargs) -> Driver:
         return GraphDatabase.driver(


neo4j/testcontainers/neo4j/__init__.py
```diff
@@ -76,8 +76,7 @@ def _connect(self) -> None:
         with self.get_driver() as driver:
             # Drivers may or may not be lazy
             # force them to do a round trip to confirm neo4j is working
-            with driver.session() as session:
-                session.run("RETURN 1").single()
+            driver.verify_connectivity()
```
tillahoffmann, line 79: Neat! Do we need to add the `DriverError` raised by `verify_connectivity` to the list of transient exceptions (see the definition of `wait_container_is_ready`)?

robsdedude: The errors that `verify_connectivity` can raise should be a subset (not sure if strict) of what the dummy query would have raised. So if what you had worked well, this change should not require changing the error handling around it. I sifted through the testcontainer's retry mechanism an am not 100% sure I understand what errors you want to catch. So please take the following with a grain of salt:

You might be looking for `is_retryable` which is defined on `DriverError` ([docs](https://neo4j.com/docs/api/python-driver/5.5/api.html#neo4j.exceptions.DriverError.is_retryable)) and `Neo4jError` ([docs](https://neo4j.com/docs/api/python-driver/5.5/api.html#neo4j.exceptions.Neo4jError.is_retryable)).  
A little extra context: `DriverError` is used for everything that goes wrong on the client side (before being able to connect to the DBMS, e.g., no connection, not able to fetch routing information, misconfigured/misused driver, ...) and `Neo4jError` is used for any error the server returns.

What the API docs don't mention (I'll fix that soon): `is_retryable` was added in driver version 5.0. Checking https://github.com/testcontainers/testcontainers-python/blob/master/neo4j/setup.py, I don't see any bound on the driver version. So I'm not sure you can rely on that.

tillahoffmann: I see, we already check whether the container is up with the logs so if this check fails... it should fail. Thanks for cleaning this up.

PR #320, opened 2023-03-13
add support async postgres driver

fourteekey:


postgres/testcontainers/postgres/__init__.py
```diff
@@ -59,9 +60,12 @@ def _configure(self) -> None:
         self.with_env("POSTGRES_PASSWORD", self.POSTGRES_PASSWORD)
         self.with_env("POSTGRES_DB", self.POSTGRES_DB)
 
-    def get_connection_url(self, host=None) -> str:
+    def get_connection_url(self, host=None, driver: str = None) -> str:
```
pffijt, line 63: Why does someone want to supply driver when supplied in constructor?

core/testcontainers/core/generic.py
```diff
@@ -28,13 +28,16 @@ class DbContainer(DockerContainer):
...
     @wait_container_is_ready(*ADDITIONAL_TRANSIENT_ERRORS)
-    def _connect(self) -> None:
+    def _connect(self, driver: Optional[str] = None) -> None:
         import sqlalchemy
-        engine = sqlalchemy.create_engine(self.get_connection_url())
+        engine = sqlalchemy.create_engine(self.get_connection_url(driver=driver))
         engine.connect()
 
-    def get_connection_url(self) -> str:
+    def get_connection_url(self, host=None, driver=None) -> str:
```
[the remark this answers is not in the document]
pffijt: A change just at PostgresqlContainer is fine for now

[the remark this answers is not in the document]
fourteekey: ok

fourteekey:
@pffijt 
Do you have any questions or comments about commits?

postgres/setup.py
```diff
@@ -14,6 +14,8 @@
         "testcontainers-core",
         "sqlalchemy",
         "psycopg2-binary",
+        "asyncpg",
+        "pytest-asyncio",
```
tillahoffmann, line 18: Let's move this to the [`requirements.in`](https://github.com/testcontainers/testcontainers-python/blob/master/requirements.in) because `pytest-asyncio` is only required for testing.

postgres/testcontainers/postgres/__init__.py
```diff
@@ -41,10 +49,14 @@ class PostgresContainer(DbContainer):
     POSTGRES_USER = os.environ.get("POSTGRES_USER", "test")
     POSTGRES_PASSWORD = os.environ.get("POSTGRES_PASSWORD", "test")
     POSTGRES_DB = os.environ.get("POSTGRES_DB", "test")
+    DEFAULT_DRIVER = "psycopg2"
 
     def __init__(self, image: str = "postgres:latest", port: int = 5432, user: Optional[str] = None,
                  password: Optional[str] = None, dbname: Optional[str] = None,
-                 driver: str = "psycopg2", **kwargs) -> None:
```
tillahoffmann, line 47: What's the motivation for moving the default value to a class variable?

postgres/testcontainers/postgres/__init__.py
```diff
@@ -54,14 +66,24 @@ def __init__(self, image: str = "postgres:latest", port: int = 5432, user: Optio
 
         self.with_exposed_ports(self.port_to_expose)
 
+    @wait_container_is_ready(*ADDITIONAL_TRANSIENT_ERRORS)
+    def _connect(self) -> None:
+        import sqlalchemy
+        engine = sqlalchemy.create_engine(self.get_connection_url(driver=self.DEFAULT_DRIVER))
```
tillahoffmann, line 72: Can we just use the `super` call here? It looks like the method body is the same as in the parent class (except for the driver argument which is already handled in the updated `get_connection_url`).

tillahoffmann (changes requested):
Thanks for the contribution. Added a few inline questions.

tillahoffmann:
Unresolved some of the outstanding questions, @fourteekey. We'll also have to regenerate the lock files in `requirements.txt`. See step three of https://github.com/testcontainers/testcontainers-python#contributing-a-new-feature for details.

fourteekey:
@tillahoffmann 

All questions have been resolved.  
I can't regenerate the lock files without approve,  I got error: `RuntimeError: Could not find a workflow. Has the GitHub Action run completed? If youare a first-time contributor, a contributor has to approve your changesbefore Actions can run.`

postgres/testcontainers/postgres/__init__.py
tillahoffmann: @fourteekey, could you please address this question?

alexanderankin:
closing in favor of #445 because i believe it will handle this use case, let me know if this is not the case

PR #330, opened 2023-04-12
Release 4.0.0.

tillahoffmann:
Does what it says on the box and adds a note to the documentation and bug issue template.

PR #337, opened 2023-04-19
feat(compose): allow running specific services in compose

### LINKED ISSUE #315: Docker compose start with selected service
Hello there, I wonder if we could run the docker container based on a docker compose file, but only run the selected service inside it? like doing this command `docker-compose up [service_name]`, is there any method can support that?

Currently we're using:
```
import testcontainers.compose
compose = testcontainers.compose.DockerCompose(COMPOSE_PATH)
compose.start()
```
But it will start all the services, but we just want to start selected services.

Thanks!!

balint-backmaker:
# Changes

- the `DockerCompose` class now has a new optional argument for `services` 
- this allows you to run specific services only instead of all of them
- auto-sorting of imports in the `compose/__init__.py` for convenience (no proper `isort` right now, just this file)

# Issues
Should resolve #315 on a very basic level. 

I'll be raising more PRs for QoL:
https://github.com/testcontainers/testcontainers-python/issues?q=is%3Aopen+is%3Aissue+compose


compose/testcontainers/compose/__init__.py
```diff
@@ -1,9 +1,10 @@
-import requests
 import subprocess
 from typing import Iterable, List, Optional, Tuple, Union
 
-from testcontainers.core.waiting_utils import wait_container_is_ready
+import requests
+
 from testcontainers.core.exceptions import NoSuchPortExposed
+from testcontainers.core.waiting_utils import wait_container_is_ready
```
balint-backmaker, line 7: auto sort:
- system deps
- std lib deps
- interanl deps

compose/tests/test_docker_compose.py
```diff
@@ -40,6 +39,19 @@ def test_can_build_images_before_spawning_service_via_compose():
...
+    with patch.object(DockerCompose, "_call_command") as call_mock:
+        with DockerCompose(ROOT, services=["hub", "firefox"]) as compose:
+            ...
+
+    assert compose.services
+    docker_compose_cmd = call_mock.call_args_list[0][1]["cmd"]
+    services_at_the_end = docker_compose_cmd[-2:]
+    assert "firefox" in services_at_the_end
+    assert "hub" in services_at_the_end
+    assert "chrome" not in docker_compose_cmd
```
balint-backmaker, line 52: for this one, I replicated the test above and adjusted for my needs. not mega-happy with it, but should do the job for now. I'll be sure to come back and clean this test file up a bit on future contributions.

balint-backmaker (commented):
adding author annotations for easier understanding of my intent.

compose/tests/test_docker_compose.py
tillahoffmann: Can we check whether the behavior is as expected? E.g., run only the `firefox` service and check that it works but the `chrome` service does not work. And then the reverse?

balint-backmaker:
@tillahoffmann I've updated the tests to exercise the `docker-compose`. Unfortunately couldn't do it easily on `chrome` or `firefox` because they exit pretty fast and don't want to understand the container mechanics right off the bat.

As a workaround I ended up exercising the `hub` and thus turning it on/off and proving my point that way.
With Docker Compose V2, this test will become better asserted so I think I'll go there next after this one. Yay for `docker-compose ps --format json` :)

tillahoffmann (approved):

tillahoffmann:
Looks good; just a linting error to fix.

balint-backmaker:
> Looks good; just a linting error to fix.

fixed - can you rerun the tests for me and merge at will?

I've started looking at the compose-v2 stuff but I'll move that onto the issue discussion (because v2 isn't a python package any more so rather would stick it to the issue instead of a PR)

compose/testcontainers/compose/__init__.py
```diff
@@ -16,6 +17,7 @@ class DockerCompose:
         pull: Pull images before launching environment.
         build: Build images referenced in the configuration file.
         env_file: Path to an env file containing environment variables to pass to docker compose.
+        services: The list of services to be run when starting this DockerCompose.
```
tillahoffmann, line 20:
```suggestion
        services: List of services to start.
```

tillahoffmann (approved):

MCatherine1994:
Hello there, I wonder if this feature is already included? or we still need to wait for a new version release?

Thanks!!

tillahoffmann:
Not yet, but should be soon when #330 is merged.

PR #341, opened 2023-04-27
fix: Update the copyright header for readthedocs

oleg-nenashev:
The current one references 2017, and it is confusing since the website may be perceived as outdated one


conf.py
```diff
@@ -54,7 +54,7 @@
 
 # General information about the project.
 project = u'testcontainers'
-copyright = u'2017, Sergey Pirogov'
+copyright = u'2017-2023, Sergey Pirogov and Testcontainers Python contributors'
```
tillahoffmann, line 57: Thoughts on "since 2017" so we don't need to update again?

barjb: Changes are fine. Convenction with current year in the description leaves best impression on new users.

barjb (approved):

PR #347, opened 2023-05-06
Add macos build.

tillahoffmann:
Tests seem to fail sometimes because Docker doesn't start up in time. https://github.com/actions/runner-images/issues/1143#issuecomment-652264388 may be an alternative setup option.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,6 +20,8 @@ jobs:
           python-version: "3.10"
         - machine: ubuntu-latest
           python-version: "3.11"
+        - machine: macos-latest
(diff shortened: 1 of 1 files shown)


santi:
Closed due to hardware virtualization limits on MacOS M1 runners by Github. https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/about-larger-runners#limitations-for-macos-larger-runners

Until a runner with M2+ architecture is released or another ARM64-runner is available, testing on ARM64 is not possible without heavy virtualization, which would make the test suite take forever to complete.

Also related to https://github.com/testcontainers/testcontainers-python/pull/520

Issue #289, opened 2022-12-28
KeycloakContainer image


expploitt:
Hi! I was trying to change from ` jboss/keycloak` image to `quay.io/keycloak/keycloak:latest` because the jboss repo is not longer supported. However, I don't know why, I'm getting TimeoutException during port mapping for container.

Here I have the logs:

```
============================= test session starts ==============================
collecting ... collected 1 item
run-last-failure: rerun previous 1 failure

test_security_controller.py::TestSecurityController::test_auth_token FAILED [100%]Pulling image quay.io/keycloak/keycloak:latest
Container started: 4f0e9d085c71
Waiting to be ready...
Waiting to be ready...
Waiting to be ready...

eees_eas_discovery/test/test_security_controller.py:8 (TestSecurityController.test_auth_token)
self = <eees_eas_discovery.test.test_security_controller.TestSecurityController testMethod=test_auth_token>

    def test_auth_token(self):
        keycloak_container = KeycloakContainer()
    
>       with KeycloakContainer(image='quay.io/keycloak/keycloak:latest') as kc:

test_security_controller.py:12: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../venv/lib/python3.8/site-packages/testcontainers/core/container.py:71: in __enter__
    return self.start()
../../venv/lib/python3.8/site-packages/testcontainers/keycloak.py:61: in start
    self._connect()
../../venv/lib/python3.8/site-packages/testcontainers/core/waiting_utils.py:49: in wrapper
    return wrapped(*args, **kwargs)
../../venv/lib/python3.8/site-packages/testcontainers/keycloak.py:54: in _connect
    url = self.get_url()
../../venv/lib/python3.8/site-packages/testcontainers/keycloak.py:49: in get_url
    port = self.get_exposed_port(self.port_to_expose)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

wrapped = <bound method DockerContainer.get_exposed_port of <testcontainers.keycloak.KeycloakContainer object at 0x7fe0ae54a310>>
instance = <testcontainers.keycloak.KeycloakContainer object at 0x7fe0ae54a310>
args = (8080,), kwargs = {}

[...]

PR #357, opened 2023-05-23
Fix tests on Apple Silicon / M1

santi:
Updates some tests to make them compatible with M1 / Apple Silicon Macs. Also adds some `wait_for_logs` utilities in tests to avoid race conditions on images with slow startup time (Might be more race conditions in other tests, but these were the ones i encountered myself). 

Quick summary of changes:
- Removes ElasticSearch v6.x, which is no longer supported)
- Updates nginx image to support ARM64/v8, and waits for logs before running asserts.
- Updates ArangoDB to latest stable release, which supports ARM64/v8.
- Pins the Redis image and version to avoid using `:latest` in tests, for more reproducible builds. Also waits for logs before connecting.

Only remaining failing test on Apple Silicon is the `jboss/keycloak` image, which seems to be unmaintained (latest release 1 year ago+ and the images have officially moved to `quay.io`. Should we remove this altogether as part of the 4.0 release of `testcontainers`?

Update: Updated `keycloak` image to new, maintained repo on `quay.io`. Works great on Apple Silicon and other architectures.

### DIFF
--- a/keycloak/testcontainers/keycloak/__init__.py
+++ b/keycloak/testcontainers/keycloak/__init__.py
@@ -43,6 +43,7 @@ class KeycloakContainer(DockerContainer):
     def _configure(self) -> None:
         self.with_env("KEYCLOAK_ADMIN", self.username)
         self.with_env("KEYCLOAK_ADMIN_PASSWORD", self.password)
+        self.with_command("start-dev")
 
     def get_url(self) -> str:
         host = self.get_container_host_ip()
--- a/keycloak/testcontainers/keycloak/__init__.py
+++ b/keycloak/testcontainers/keycloak/__init__.py
@@ -11,12 +11,11 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 import os
-import requests
 
 from keycloak import KeycloakAdmin
 
 from testcontainers.core.container import DockerContainer
-from testcontainers.core.waiting_utils import wait_container_is_ready
+from testcontainers.core.waiting_utils import wait_for_logs
 from typing import Optional
 
 
@@ -33,45 +32,32 @@ class KeycloakContainer(DockerContainer):
             >>> with KeycloakContainer() as kc:
             ...    keycloak = kc.get_client()
     """
-    def __init__(self, image="jboss/keycloak:latest", username: Optional[str] = None,
+    def __init__(self, image="quay.io/keycloak/keycloak:latest", username: Optional[str] = None,
                  password: Optional[str] = None, port: int = 8080) -> None:
         super(KeycloakContainer, self).__init__(image=image)
-        self.username = username or os.environ.get("KEYCLOAK_USER", "test")
-        self.password = password or os.environ.get("KEYCLOAK_PASSWORD", "test")
+        self.username = username or os.environ.get("KEYCLOAK_ADMIN", "test")
+        self.password = password or os.environ.get("KEYCLOAK_ADMIN_PASSWORD", "test")
         self.port = port
         self.with_exposed_ports(self.port)
 
     def _configure(self) -> None:
-        self.with_env("KEYCLOAK_USER", self.username)
-        self.with_env("KEYCLOAK_PASSWORD", self.password)
+        self.with_env("KEYCLOAK_ADMIN", self.username)
+        self.with_env("KEYCLOAK_ADMIN_PASSWORD", self.password)
 
     def get_url(self) -> str:
         host = self.get_container_host_ip()
         port = self.get_exposed_port(self.port)
         return f"http://{host}:{port}"
 
-    @wait_container_is_ready(requests.exceptions.ConnectionError, requests.exceptions.ReadTimeout)
-    def _connect(self) -> None:
-        url = self.get_url()
-        response = requests.get(f"{url}/auth", timeout=1)
-        response.raise_for_status()
-
     def start(self) -> "KeycloakContainer":
         self._configure()
-        super().start()
-        self._connect()
+        container = super().start()
+        wait_for_logs(container, 'Listening on:')
         return self
 
     def get_client(self, **kwargs) -> KeycloakAdmin:
-        default_kwargs = dict(
-            server_url=f"{self.get_url()}/auth/",
+        return KeycloakAdmin(
+            server_url=f"{self.get_url()}/",
             username=self.username,
             password=self.password,
-            realm_name="master",
-            verify=True,
-        )
-        kwargs = {
-            **default_kwargs,
-            **kwargs
-        }
-        return KeycloakAdmin(**kwargs)
+            verify=True)
--- a/nginx/tests/test_nginx.py
+++ b/nginx/tests/test_nginx.py
@@ -1,12 +1,10 @@
 import requests
-from testcontainers.core.waiting_utils import wait_for_logs
 
 from testcontainers.nginx import NginxContainer
 
 
 def test_docker_run_nginx():
     with NginxContainer("nginx:1.24.0") as nginx:
-        wait_for_logs(nginx, 'ready for start up')
         url = f"http://{nginx.get_container_host_ip()}:{nginx.get_exposed_port(nginx.port)}/"
         r = requests.get(url)
         assert (r.status_code == 200)
--- a/nginx/tests/test_nginx.py
+++ b/nginx/tests/test_nginx.py
@@ -1,11 +1,12 @@
 import requests
+from testcontainers.core.waiting_utils import wait_for_logs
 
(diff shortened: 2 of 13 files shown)


nginx/tests/test_nginx.py
```diff
@@ -1,11 +1,12 @@
...
+from testcontainers.core.waiting_utils import wait_for_logs
 
 from testcontainers.nginx import NginxContainer
 
 
 def test_docker_run_nginx():
-    nginx_container = NginxContainer("nginx:1.13.8")
-    with nginx_container as nginx:
+    with NginxContainer("nginx:1.24.0") as nginx:
+        wait_for_logs(nginx, 'ready for start up')
```
tillahoffmann, line 9: Great! Is this something we should move into the container implementation?

redis/tests/test_redis.py
```diff
@@ -1,23 +1,24 @@
...
+from testcontainers.core.waiting_utils import wait_for_logs
 
 from testcontainers.redis import RedisContainer
 
 
 def test_docker_run_redis():
-    config = RedisContainer()
-    with config as redis:
+    with RedisContainer('redis:7.0.11') as redis:
+        wait_for_logs(redis, 'Ready to accept connections')
```
tillahoffmann, line 9: Same, should we move this into the container implementation?

redis/tests/test_redis.py
```diff
@@ -1,23 +1,24 @@
...
+        msg = wait_for_message(pubsub)
         assert 'data' in msg
         assert b'new_msg', msg['data']
 
 
 def test_docker_run_redis_with_password():
-    config = RedisContainer(password="mypass")
-    with config as redis:
+    with RedisContainer('redis:7.0.11', password="mypass") as redis:
+        wait_for_logs(redis, 'Ready to accept connections')
```
tillahoffmann, line 21: Same, should we move this into the container implementation?

tillahoffmann (commented):
Thanks for the fixes; left a few inline comments.

Regarding keycloack, there's a possible alternative image discussed in #289, but we don't have a PR yet.

redis/tests/test_redis.py
santi: Maybe. It depends on how we want to support these "extra" containers, as previous versions of `nginx`, `redis` etc does not necessarily have the same logs (Haven't checked for `redis`, but `nginx` is completely silent on startup up until v1.20).

If we want to only support modern versions it would be okay, and in that case we should explicitly list the version range we support. IMO this would be okay, as I believe most users only want to run a modern instance of the image, and the ones that specifically want an older version can instead specify that as its own `DockerContainer`. 

So yeah, I think we should do it and update the docs to indicate it.

redis/tests/test_redis.py
santi: > Regarding keycloack, there's a possible alternative image discussed in #289, but we don't have a PR yet.
I tried for a few minutes to get the maintained image up and running, but at first sight it looked like it is intended that you specify at least an `ENTRYPOINT` while using the image as a base image. E.g. it is not 100% Plug'n'Play. I might be wrong though, I did not give myself too much time trying to fix it. 

I might have some time to look into it at a later stage, but I really want to get `.with_reuse()` up and running after the Ryuk PR is merged. Might be that it annoys me enough during development that I cannot get all the tests to run on my machine 😅

santi:
> Regarding keycloack, there's a possible alternative image discussed in https://github.com/testcontainers/testcontainers-python/issues/289, but we don't have a PR yet.

I tried for a few minutes to get the maintained image up and running, but at first sight it looked like it is intended that you specify at least an ENTRYPOINT while using the image as a base image. E.g. it is not 100% Plug'n'Play. I might be wrong though, I did not give myself too much time trying to fix it.

I might have some time to look into it at a later stage, but I really want to get .with_reuse() up and running after the Ryuk PR is merged. Might be that it annoys me enough during development that I cannot get all the tests to run on my machine 😅

redis/tests/test_redis.py
tillahoffmann: That sounds like a good plan.

nginx/tests/test_nginx.py
santi: Moved logic to be part of the `start` method 👍

redis/tests/test_redis.py
santi: Moved logic to be part of the `start` method 👍 Updated docs along with it

keycloak/testcontainers/keycloak/__init__.py
```diff
@@ -33,45 +32,33 @@ class KeycloakContainer(DockerContainer):
...
-        url = self.get_url()
-        response = requests.get(f"{url}/auth", timeout=1)
-        response.raise_for_status()
-
     def start(self) -> "KeycloakContainer":
         self._configure()
-        super().start()
-        self._connect()
+        container = super().start()
+        wait_for_logs(container, 'Listening on:')
```
TheForgottened, line 56: I haven't tried this yet, but by passing `KC_HEALTH_ENABLED=true` as a environment variable I believe it might be possible to keep the `self._connect()` if instead of pooling `{url}/auth` we pool `{url}/health`.

TheForgottened:
Hi, just a general comment. Because of the fact that Keycloak changed their architecture (if you look at the tags, there are a bunch of legacy versions) there's no silver bullet to spin up a container and make sure it's up and running. After not a huge time I was able to come up with this solution, it's not perfect but from my testing it seems to be working for all Keycloak versions, including legacy and non legacy ones.

```python
class CustomKeycloakContainer(DockerContainer):
    def __init__(
            self,
            image="quay.io/keycloak/keycloak:latest",
            username: Optional[str] = None,
            password: Optional[str] = None,
            port: int = 8080
    ) -> None:
        super(CustomKeycloakContainer, self).__init__(image=image)
        self.username = username or os.getenv("KEYCLOAK_USER", "test")
        self.password = password or os.getenv("KEYCLOAK_PASSWORD", "test")
        self.port = port
        self.with_exposed_ports(self.port)

    def _get_image_tag(self) -> str:
        return self.image.split(":")[1]

    def _uses_legacy_architecture(self) -> bool:
        tag = self._get_image_tag()

        try:
            major_version_number = int(tag.split(".")[0])
        except ValueError:
            major_version_number = 2023

        return major_version_number <= 16 or "legacy" in tag.lower()

    def _configure(self) -> None:
        # There's a bit of confusion as to which version / architecture uses which
        # Because of this I'm setting the credentials both ways
        self.with_env("KEYCLOAK_ADMIN", self.username)
        self.with_env("KEYCLOAK_ADMIN_PASSWORD", self.password)
        self.with_env("KEYCLOAK_USER", self.username)
        self.with_env("KEYCLOAK_PASSWORD", self.password)

        self.with_env("KC_HEALTH_ENABLED", "true")
        self.with_env("KC_METRICS_ENABLED", "true")

        if not self._uses_legacy_architecture():
            self.with_command("start-dev")

    def get_url(self) -> str:
        host = self.get_container_host_ip()
        port = self.get_exposed_port(self.port)
        return f"http://{host}:{port}"

    def get_base_api_url(self) -> str:
        base_url = self.get_url()
        return f"{base_url}/auth" if self._uses_legacy_architecture() else base_url

    @wait_container_is_ready(requests.exceptions.ConnectionError, requests.exceptions.ReadTimeout)
    def _connect(self) -> None:
        url = self.get_url()
        response = requests.get(f"{url}/auth", timeout=1)
        response.raise_for_status()

    def start(self) -> "CustomKeycloakContainer":
        self._configure()
        container = super().start()

        if self._uses_legacy_architecture():
            self._connect()
        else:
            wait_for_logs(container, 'Listening on:')

        return self
```

This might need some refinement but overall I'm happy with the solution. This is something hard to tackle because of all the changes between versions and architectures and the lack of decent documentation.

santi:
I think we should have a larger discussion in this repo on what we actually want to support in custom container setups, as making the implementation support all versions since the beginning of time will lead to a lot of nasty workarounds while only catering to a (presumably) small portion of people that want to use legacy versions. 

We need to ask ourselves the following question, and probably a lot more:
- Should we support the latest version?
- Should we only support officially supported versions of the service?
- What repositories should we support? What if there are changes between the images in the different repositories?
- If there is no officially supported version range, what do we do?

Having support for an explicit version range would also require us to cap the support on the latest release, and check / update the version range on every new release (otherwise we cannot know if the new versions are actually compatible or not, and it would be worse to say incorrectly that we support something than not supporting it).

I'm making an issue where we can follow this discussion more closely.

santi:
Closing due to fixes already implemented in other branches

PR #312, opened 2023-02-24
Update compose to work with both "docker-compose" as well as "docker compose" 

jhnnsrs:
Fixes #306

Sorry for the deletion of the old pull-request.

Issue #358, opened 2023-05-24
Migrate to DockerCompose V2


balint-backmaker:
# Motivation
- Docker Compose v2 offers more features that we can make use of and create a better developer experience.
- Docker Compose v1 will no longer be receiving updates after June 2023 so we should deprecate the usage of it

## Notable features
- `docker compose ps --format json` can help us shortcut the effort on many useful things
- #226 is trivial after this and fairly safe to "prefer the plugin over the standalone"
- #259 may have been fixed in v2 but any fixes might be better implemented for v2
- #241 is more trivial to implement once we have the json format available to us
- #239 is also trivial as well as giving us a much richer set of data, including the easy shortcut to get the container ID and allow `testcontainers` to start/stop these containers

# The challenges
- dual support of v1 and v2: 
  - legacy support means tougher feature parity and more testing needed for something that is in fact deprecated 
  - I'd like to drop it, but this means compose contains a breaking change
  - so what is the next major release for `testcontainers-python` ? :slightly_smiling_face: 
- docker compose v2 is not a python package (it's written in golang and I'm not sure I'm up for making a binding for this)
  - so need to agree on a sensible way to check this on instantiation of a `DockerCompose` object. 
  - Not hard, just needs some reviews and is a breaking change

# Final thoughts
I'm happy to pick this up, in fact I have a local copy with a bunch of goodies, just need some community support in how to introduce this change.

Also since this is likely a breaking change for the compose plugin, your thoughts on how to version this would be very welcome @tillahoffmann :pray:

antoine-de:
This would be a really nice feature now that Docker compose v1 is EOL :pray:

alexanderankin:
![ContainerisedDockerCompose](https://github.com/testcontainers/testcontainers-python/assets/8921892/c53949d7-8065-41e5-8f98-4db177f4965d)
here is how testcontainers-java does it

e.g. https://github.com/testcontainers/testcontainers-java/blob/main/core/src/main/java/org/testcontainers/containers/DockerCompose.java

alexanderankin:
merged in #426 , released in #330

PR #362, opened 2023-06-05
Add support for docker compose v2

### LINKED ISSUE #306: Allow configuring the `docker-compose` executable path
**Describe the bug**

On newer docker setups, docker-compose is part of docker under a docker sub commands: `docker compose`

This makes this library fail to find `docker-compose` when needed.

For now I have a little script but I expect this library to be able to change the executable name:
```sh
#!/usr/bin/env bash

exec docker compose "$@"
```

psschwei:
Co-authored-by: Johannes Roos jhnnsrs@gmail.com

Fixes #306 

Uses @jhnnsrs 's approach in #312 while refactoring to use the new structure post #290 
Also tweaked the tests a bit so they would pass when using `docker compose`


psschwei:
Seems `cached_property` isn't implemented in Python 3.7, which is what's causing those tests to fail.

compose/testcontainers/compose/__init__.py
```diff
@@ -65,14 +70,33 @@ def __enter__(self) -> "DockerCompose":
...
+            The docker compose command parts
+        """
+        if self._user_defined_compose_command:
+            return self._user_defined_compose_command
+
+        if subprocess.run(["docker", "compose", "--help"], stdout=subprocess.DEVNULL,
+                          stderr=subprocess.STDOUT).returncode == 0:
+            return ["docker", "compose"]
+
+        return ["docker-compose"]
```
balint-backmaker, line 90: can you run this once on the constructor? it would save some noise and confusion and it's better to fail early.

balint-backmaker (commented):
Nice! Glad to see more people looking at this! :heart: 

I've got one small request but I'm no `codeowner` so no "request changes")

compose/testcontainers/compose/__init__.py
psschwei: good idea!

psschwei:
@tillahoffmann a few of the tests were canceled, but I didn't see any errors in the logs... is there anything else needed for this one?

compose/testcontainers/compose/__init__.py
```diff
@@ -57,6 +61,8 @@ def __init__(
         self.build = build
         self.env_file = env_file
         self.services = services
+        self._user_defined_compose_command = compose_command.split(" ") if compose_command else None
+        self.docker_compose = self.base_docker_compose()
```
tillahoffmann, line 65: Can we clean up the naming conventions here? We currently have

- `_user_defined_compose_command` as a private attribute.
- `compose_command` as a parameter of the constructor.
- `docker_compose` as a public attribute.
- `base_docker_compose` as a public function.

Can we consolidate?

compose/testcontainers/compose/__init__.py
```diff
@@ -65,14 +71,33 @@ def __enter__(self) -> "DockerCompose":
     def __exit__(self, exc_type, exc_val, exc_tb) -> None:
         self.stop()
 
+    def base_docker_compose(self):
```
tillahoffmann, line 74: Let's make this a private method starting with `_` because we don't need to expose it to the user.

tillahoffmann:
> @tillahoffmann a few of the tests were canceled, but I didn't see any errors in the logs... is there anything else needed for this one?

They were cancelled because the tests timed out after a few hours. There are likely some issues with spinning up the containers.

compose/testcontainers/compose/__init__.py
psschwei: dropped `_user_defined_compose_command`, made `base_docker_compose` a private function (and renamed), and renamed `docker_compose` attribute to `compose_command` to mirror the constructor

psschwei:
looks like I may need to add a comment to get the "requires attention" label back :smile:

balint-backmaker:
Hey folks! I know things may be going slow right now, but I'd love to put some effort into #358 and would love to do so after this PR has been merged (likely to overhaul the compose package, but this contains some good ideas I want to keep)

psschwei:
@tillahoffmann gentle ping

psschwei:
The compose module was removed in #394 , so doesn't seem there's any need for this PR anymore...

kiview:
As said in https://github.com/testcontainers/testcontainers-python/pull/394#issuecomment-1826006889, we are interested in bringing back support to Compose v2 without depending on the Python Docker Compose v1 dependency (so only delegating to the binary). Can this PR be adopted towards this approach?

psschwei:
I understand the rationale for dropping the compose module. No hard feelings there :smile: 

As for adapting the PR... that's a possibility, though I'm not sure if/when I'd have the cycles... so and rather leaving the PR open indefinitely, I figured I'd go ahead and close it. If someone else wanted to resurrect it, that'd be okay with me too.

Issue #305, opened 2023-02-09
Mark packages as typed


jooola:
**Describe the bug**

Mypy is failing because the packages aren't marked as typed (missing `py.typed` file), but it seems that most of the packages have the type definitions.

Could the packages be marked as typed ?

**To Reproduce**

```python
from testcontainers.compose import DockerCompose

with DockerCompose("."):
  pass
```

```
$ mypy tests
tests/e2e_test.py:12: error: Skipping analyzing "testcontainers.compose": module is installed, but missing library stubs or py.typed marker  [import]
tests/e2e_test.py:12: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports

```

tillahoffmann:
Yes, that should definitely be possible. Would you be up for setting up a PR? We could also add mypy checks in the same PR.

max-pfeiffer:
@totallyzen @alexanderankin I would like to pick up this one. Can you assign it to me or how does it work in this project?

alexanderankin:
@jooola if you have any feedback about #504 it is welcome!

brandur:
Does anyone know off hand whether this one was ever resolved?

I'm kind of a Python noob and was trying to add type checking on a project's tests. I got almost all of it working, except for testcontainers:

```
tests/conftest.py:5: error: Skipping analyzing "testcontainers.postgres": module is installed, but missing library stubs or py.typed marker  [import-untyped]
```

Looking in the file I'm using, it is typed, but I think the problem is that it's not declaring itself as such. Fixing this might be as simple as adding a `py.typed` file to the project.

alexanderankin:
there is a pr here but its not a top priority: #504 - almost immediately after i opened there were two pr's with minor typing fixes that conflicted with it, and i did not rebase in time. at this point it would be not only harder to rebase but also there are new modules that need types/mypy checking.

contributions would be accepted for it.

brandur:
Thanks. My Python skills are such that I'm already having enough trouble just wrangling types in one much smaller package so I don't think I'll try my hand at this yet, but it's good to know I'm not missing something.

In case this helps anyone else, here's a block that can be placed in `mypy.ini` to flag it as ignored so that MyPy can pass for everything else:

``` ini
# testcontainers is typed, but it doesn't correctly declare itself as such.
# Hopefully it can be fixed one day:
#
# https://github.com/testcontainers/testcontainers-python/issues/305
[mypy-testcontainers.*]
[...]

PR #365, opened 2023-07-02
[Issue 305] Mark packages as typed

RayanRal:
Added py.typed to all modules - empty file, just making types "discoverable" by mypy
Should resolve https://github.com/testcontainers/testcontainers-python/issues/305


alexanderankin:
this particular implementation is no longer relevant (going to tag #305 though as mypy is still on the table)

PR #366, opened 2023-07-04
fix(vault): add support for HashiCorp Vault container

f4z3r:
Add support for a Vault container.


f4z3r:
@tillahoffmann Am I correct in assuming I need to wait for workflow approval before I can continue to update the dependencies for this PR?

santi:
Hi @f4z3r!

Thank you for your contribution, Hashicorp's Vault is something we would like to see supported. We are sorry for the extended delay in feedback and merged PRs, but are now on back on track with active maintainers.

Could you please rebase or merge in the latest changes from `main`? That way we can start moving this PR forward

f4z3r:
@santi alright I updated the PR to be according to the new repo structure. I am not super familiar with Poetry, but it seemed I needed to update the entire lock file to install the required packages I added to run the tests. As far as I could check, it seems to run fine, `make modules/vault/tests` ran the tests successfully.

modules/vault/tests/test_vault.py
```diff
@@ -0,0 +1,39 @@
+from testcontainers.vault import VaultContainer
+
+
+def test_docker_run_vault():
+    config = VaultContainer()
```
santi, line 5: For our tests we want to pin the vault version to a specific tag in order to have reproducible tests. Could you pin the image version to `VaultContainer('hashicorp/vault:1.16.1')`? 

The `latest` tag in the default constructor argument can stay as-is.

poetry.lock
santi, line 1: Judging by the large diff in this file, I suspect you might have performed a `poetry lock`, without the `--no-update` flag. This updates all dependency versions and locks them, not just the ones you have added. Could you reset the lockfile to the one found in the current `main` branch and run `poetry lock --no-update` instead? That way we only change the libraries you added

modules/vault/testcontainers/vault/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+        Returns:
+            client: Vault client to connect to the container.
+        """
+        config = self.get_config()
+        return hvac.Client(url=config["address"], token=config["root_token"])
+
+    @wait_container_is_ready(requests.ConnectionError)
+    def _healthcheck(self) -> None:
+        url = f"{self.get_config()['address']}/v1/sys/health"
+        response = requests.get(url, timeout=3)
```
santi, line 81: We are trying to heavily cut down on dependencies for this package, and would like to not bundle `requests` as part of this lib. Could you rewrite to using the built-in `urllib` instead? Something like 
```
from urllib.request import urlopen
from http.client import HTTPException

@wait_container_is_ready(HTTPException)
def _healthcheck(self) -> None:
    url = f"{self.get_config()['address']}/v1/sys/health"
    with urlopen(url) as res:
        if res.status > 299:
            raise HTTPException()
```

modules/vault/testcontainers/vault/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+    Vault container.
+
+    Example:
+
+        .. doctest::
+
+            >>> from testcontainers.vault import VaultContainer
+
+            >>> with VaultContainer() as vault_container:
```
santi, line 31: Could you also pin the dependency version here as well? These examples are run as tests, so it's nice to have the version pinned in them as well

modules/vault/testcontainers/vault/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+        """
+        Get a Vault client.
+
+        Returns:
+            client: Vault client to connect to the container.
+        """
+        config = self.get_config()
+        return hvac.Client(url=config["address"])
+
+    def get_root_client(self) -> hvac.Client:
```
santi, line 68: As part of a current discussion in https://github.com/testcontainers/testcontainers-python/issues/526, we try to not bundle the client library in the `testcontainers` package itself, but leave it up to the implementer to use their preferred client of choice. It is not 100% decided what the strategy should be for this, but since it is easier to add than to remove public APIs in the future, I think we should skip these two methods (`get_client` and `get_root_client`) and instead have usage of the clients documented in the `docstring`/`doctest` above.

Just add the `hvac` package to the dev dependency group (`poetry add "hvac==*" --group dev`) so that it is available for tests

modules/vault/testcontainers/vault/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+    def __init__(self, image: str = "hashicorp/vault:latest", port: int = 8200,
+                 root_token: str = "toor", **kwargs) -> None:
+        raise_for_deprecated_parameter(kwargs, "port_to_expose", "port")
+        super(VaultContainer, self).__init__(image, **kwargs)
+        self.port = port
+        self.root_token = root_token
+        self.with_exposed_ports(self.port)
+        self.with_env("VAULT_DEV_ROOT_TOKEN_ID", self.root_token)
+
+    def get_config(self) -> dict:
```
santi, line 44: This looks a bit Java-like to me. I saw it added in the `chroma` and `minio` modules, but I really think we instead should expose a `get_connection_url` method which gives you the `address` from the `dict` in this field. The `root_token` is available as an instance member on this object anyways, and I don;t think we need an extra method to expose it.

modules/vault/testcontainers/vault/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+        .. doctest::
+
+            >>> from testcontainers.vault import VaultContainer
+
+            >>> with VaultContainer() as vault_container:
+            ...     vault_client = vault_container.get_client()
+    """
+    def __init__(self, image: str = "hashicorp/vault:latest", port: int = 8200,
+                 root_token: str = "toor", **kwargs) -> None:
+        raise_for_deprecated_parameter(kwargs, "port_to_expose", "port")
```
santi, line 37: Since this is a new module, no need to deprecate parameters ;)

santi:
Good work so far, @f4z3r! With this speed we are ready to merge the module today 😄

modules/vault/testcontainers/vault/__init__.py
```diff
@@ -0,0 +1,64 @@
...
+    Vault container.
+
+    Example:
+
+        .. doctest::
+
+            >>> from testcontainers.vault import VaultContainer
+
+            >>> with VaultContainer("hashicorp/vault:1.16.1") as vault_container:
+            ...     connection_url = vault_container.get_connection_url()
```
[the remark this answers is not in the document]
f4z3r: Ah ok cool 👍🏽 Will do. Sorry, I am like 0 familiar with Sphinx and Poetry 😆

santi:
Tests are failing because the specified `extra` `vault` does not exist. You can add it in `pyproject.toml` under `[tool.poetry.extras]` as `vault = []` (Same as the `kafka` entry in the list)

f4z3r:
> Good work so far, @f4z3r! With this speed we are ready to merge the module today 😄

Well thanks for the quick reviews and your help! 😄

santi (approved):
Great work! 💯 Thank you for your contribution. 

Looks like Github Actions hit a bit of a bump right when everything was done, so only half of the checks have passed, and I'm getting 500 Errors randomly. Once https://github.com/testcontainers/testcontainers-python/actions/runs/8567034738, https://github.com/testcontainers/testcontainers-python/actions/runs/8567034549 and https://github.com/testcontainers/testcontainers-python/actions/runs/8567034548 have run, I'll merge 👍

f4z3r:
> Great work! 💯 Thank you for your contribution.
> 
> Looks like Github Actions hit a bit of a bump right when everything was done, so only half of the checks have passed, and I'm getting 500 Errors randomly. Once https://github.com/testcontainers/testcontainers-python/actions/runs/8567034738, https://github.com/testcontainers/testcontainers-python/actions/runs/8567034549 and https://github.com/testcontainers/testcontainers-python/actions/runs/8567034548 have run, I'll merge 👍

Yup, seems GitHub is experiencing issues: https://www.githubstatus.com/

santi:
@f4z3r Seems like the trouble is over, but the queued tasks are not running. Could you re-trigger a new set of checks by pushing a commit with `git commit --allow-empty -m "Re-trigger PR checks"`?

PR #367, opened 2023-07-05
feat(network): Add network context manager

mloesch:
This PR adds a `Network` helper class that allows to create networks and connect containers programmatically. 

The networks are context-managed resources like containers created via `DockerContainer`. 

Please also see tests for a usage example :)


SergeyPirogov (approved):

VerdantForge:
I'm interested in this feature. But I see progress has been dropped. Is there a reason?

mloesch:
> I'm interested in this feature. But I see progress has been dropped. Is there a reason?

The build is failing due to https://github.com/cython/cython/issues/4568

I created #376  to fix it, but it hasn't been approved yet.

kiview:
Hey @mloesch, thanks for providing this PR for the feature. I see you decided to allow specifying a name for the Network. In Testcontainers for Java, we made the design decision to only generate networks with random names, to avoid the potential for naming conflicts with existing networks. This is a general design approach of Testcontainers and it should be fine for the Testcontainers use cases.

Would you like to adapt the implementation accordingly?

See https://github.com/testcontainers/testcontainers-java/blob/1a3b23385f9a595576246a4c7f5dc54d244dd69d/core/src/main/java/org/testcontainers/containers/Network.java for the Java implementation.

mloesch:
Hey @kiview, thanks for the review, I changed the implementation to use random names in c10223fbec1d88755816c0ba9d2f77693816d822

core/tests/test_network.py
```diff
@@ -0,0 +1,27 @@
...
+        docker = DockerClient()
+        networks_list = docker.client.networks.list(network.name)
+        assert networks_list[0].name == network.name
+        assert networks_list[0].id == network.id
+    assert not docker.client.networks.list(network.name)
+
+
+def test_containers_can_communicate_over_network():
+    with Network() as network:
+        with DockerContainer("nginx:alpine-slim").with_name(
```
kiview, line 17: Instead of using `with_name` (which can lead to naming conflicts when parallelizing tests or accidentally reusing names), can we add an API  `with_network_aliases(string[])`, similar to [tc-java](https://java.testcontainers.org/features/networking/#advanced-networking), that uses Docker's `--network-alias` mechanism?

core/tests/test_network.py
```diff
@@ -0,0 +1,27 @@
...
+                "alpine1").with_kwargs(network=network.name) as alpine1:
```
kiview, line 18: Does it make sense to use `with_kwargs` here and not have an explicit API `with_network(Network)`?

kiview (commented):
Thanks @mloesch, that change for the random id looks good. I did another pass and compared it with the implementation in tc-java (which I consider our reference implementation) and had some more suggestions. 

I am still getting up to speed with the internals of tc-python, so we can discuss whether my suggestions make sense in this context.

core/tests/test_network.py
mloesch: :+1: added in e4d9c19

mloesch:
@kiview is there anything else required for this PR to be merged? :)

core/testcontainers/core/network.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def __enter__(self) -> "Network":
+        self._network = self._docker.client.networks.create(self.name, **self._kwargs)
+        self.id = self._network.id
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb) -> None:
+        self.remove()
+
+    def __del__(self) -> None:
```
santi, line 44: We've had some bad experiences with external resource cleanup through the `__del__` method, so we have decided to move away from this. Having the context manager and the `remove` method provides sufficient means to cleanup the resource. `__del__` hook can be removed 👍

core/tests/test_network.py
```diff
@@ -0,0 +1,38 @@
+from testcontainers.core.container import DockerContainer
+from testcontainers.core.docker_client import DockerClient
+from testcontainers.core.network import Network
+
+NGINX_ALPINE_SLIM_IMAGE = "nginx:alpine-slim"
```
santi, line 5: Can you pin this dpenendency to a specific, versioned tag? We are currently working on making our tests more reproducible, and specific version tags for images used in tests are a part of this.

santi (commented):
I went through your PR now and have some final remarks that should be addressed before we are ready to merge. Very good work so far, this is feature is awesome! 🚀

core/testcontainers/core/container.py
```diff
@@ -83,6 +94,8 @@ def start(self):
             **self._kwargs,
         )
         logger.info("Container started: %s", self._container.short_id)
+        if self._network:
+            self._network.connect(self._container.id, self._network_aliases)
```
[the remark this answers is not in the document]
mloesch: correct me if I'm wront, but `stop()` removes the container, which causes it to be removed from the network

core/testcontainers/core/container.py
```diff
@@ -42,6 +43,8 @@ def __init__(
         self._container = None
         self._command = None
         self._name = None
+        self._network: Optional[Network] = None
+        self._network_aliases: Optional[list] = None
```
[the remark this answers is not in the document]
mloesch: fixed in cac19d39176c1758ecb4878614ace3bb0d3cb6a8

core/testcontainers/core/container.py
```diff
@@ -57,6 +60,14 @@ def with_exposed_ports(self, *ports: int) -> "DockerContainer":
             self.ports[port] = None
         return self
 
+    def with_network(self, network: Network) -> "DockerContainer":
```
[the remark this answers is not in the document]
mloesch: fixed in 1c8ebdc26a42a10f2319fe127819006bd8a1f0cd

core/tests/test_network.py
mloesch: fixed in 9db1080c416c26f9cb4f542befe19dd17958ffa1

core/testcontainers/core/container.py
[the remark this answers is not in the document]
santi: Wasn't aware that `.stop()` also removed it from the network 👍

santi:
Thank you for this @mloesch! 🙌 Really appreciate your contribution, this will benefit a lot of users 💯

PR #368, opened 2023-07-05
fix(core): DinD issues #141, #329

### LINKED ISSUE #141: Testcontainers fails to get container_host_ip when running in docker network
I believe this is because this code assumes the presence of "bridge"

https://github.com/testcontainers/testcontainers-python/blob/master/testcontainers/core/docker_client.py#L52

https://github.com/testcontainers/testcontainers-python/blob/master/testcontainers/core/docker_client.py#L48

which is not the case when the container runs in a specific docker network (possible to do via DockerContainer.with_kwargs)

### LINKED ISSUE #329: Docker in docker test fails with "Error while fetching server API version"
The `core/tests/test_docker_in_docker.py::test_wait_for_logs_docker_in_docker` test has started failing in CI without an obvious reason (see [here](https://github.com/testcontainers/testcontainers-python/actions/runs/4668632702/jobs/8265981081) for an example run). https://github.com/docker/docker-py/issues/2717 may be related. @Duchadian, do you have thoughts on this?

```
FAILED core/tests/test_docker_in_docker.py::test_wait_for_logs_docker_in_docker - 
docker.errors.DockerException: Error while fetching server API version: 
HTTPConnectionPool(host='172.17.0.2', port=2375): Max retries exceeded with url: 
/version (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fda5372d090>: 
Failed to establish a new connection: [Errno 111] Connection refused'))
```

deeninetyone:
Fix #141 - find IP from custom network if the container is not using the default network
Close #329 - This seems fixed in the underlying docker libraries.
Improve support for Docker in Docker running on a custom network, by attempting to find the right custom network and use it for new containers. This adds support for using testcontainers-python running the GitHub Actions Runner Controller to run self-hosted actions runners on prem, when you run your workflows in containers.


sean-hernon:
Any movement on merging this? I am doing something similar, to index by the custom network id instead of `bridge`. Would be nice to delete my own workarounds.

damianoct:
any updates? we need to have testcontainers working properly in DIND environments

VerdantForge:
Also interested in updates on this issue

covatic-john:
Trying to get docker container in container running in bitbucket and struggling with it
I am sure that this is to do with the sock or internal host. Any pointers please

alexanderankin:
i was able to run the test suite in dind while testing compose. what does dind mean for folks here? pardon my ignorance - but want to clarify so i can understand how to manually test. i do see that there is a test here. i am adding 4.1.0 because since we have code here, we can test and merge with relatively low consequences

alexanderankin:
>  Maintainers are allowed to edit this pull request.

thank you, i will try to rebase this to get it to apply cleanly if i am able to make sense of the changes.

deeninetyone:
> > Maintainers are allowed to edit this pull request.
> 
> thank you, i will try to rebase this to get it to apply cleanly if i am able to make sense of the changes.

It fixes the case where the container you are running in has a custom network (not the default one). In this case, the library needs to try to find the correct network and join the test container to the same network, or the container you're running in cannot contact the test container.

totallyzen:
@deeninetyone thanks for the contrib! Looks promising and makes a lot of sense!

I have two asks:
- can you include `tc.host` in the "user knows best" check? It's been attempted in the [init function](https://github.com/testcontainers/testcontainers-python/blob/main/core/testcontainers/core/docker_client.py#L37) so you might want to touch that too to make sure we're doing this in a single place
- can you make the new addition more structured around functions? these nested if/for/if/for make the code hard to read. 


For the latter, my suggestion would be that you isolate the `try-catch` block into a function and instead of modifying `kwargs` from deep within the loop
- your new function would be something like `def find_host_network() -> Optional[str]:`
- then you'd do "user knows best, OR try to find host network, and if found one, alter the `kwargs`

totallyzen (approved):
LGTM, thanks for the fix! :heart:

PR #369, opened 2023-07-06
feat: Add proper support for Keycloak

TheForgottened:
The testcontainers library is giving support for a deprecated image of Keycloak.

This PR aims to give support to the new image of Keycloak, making sure it works independently of the tag used.

This has proved to be more difficult than it should've because of many reasons:

1. The lack of proper docs from Keycloak
2. Lots of differences between legacy and non-legacy Keycloak instances (with no documentation to explain better)
3. API differences between versions / architectures

The code works 100% fine according to my tests and the tests added. I haven't tested with the legacy image tbh, but I have a hunch it might just work, except for version 17 and up.

### DIFF
--- a/keycloak/testcontainers/keycloak/__init__.py
+++ b/keycloak/testcontainers/keycloak/__init__.py
@@ -11,44 +11,75 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 import os
-import requests
-
-from keycloak import KeycloakAdmin
-
-from testcontainers.core.container import DockerContainer
-from testcontainers.core.waiting_utils import wait_container_is_ready
 from typing import Optional
+
+import requests
+from keycloak import KeycloakAdmin
+from testcontainers.core.container import DockerContainer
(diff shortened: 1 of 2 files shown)


timbmg (approved):
LGTM

TheForgottened:
@timbmg anything missing on the PR that's stopping it from being merged?

maltehedderich:
@timbmg do you have the power to rerun the actions? I actually would like to use these changes, so would be nice if we could get them merged.

eddumelendez:
Thanks for your contribution, @TheForgottened . Currently, we are not able to maintain the module but there is an [Module Catalog](https://testcontainers.com/modules/), which also list community modules from [this repository](https://github.com/testcontainers/community-module-registry). I think it would be great if the repository can be hosted under https://github.com/keycloak.

/cc @stianst

disrupted:
@TheForgottened great contribution! would you be able to submit this to the Keycloak org as proposed by @eddumelendez?

Issue #379, opened 2023-08-15
DB containers produce ResourceWarning: unclosed <socket...>


sha1n:
**Describe the bug**

`DbContainer` instances may trigger a warning regarding unclosed socket. This is caused by the fact that `DbContainer._connect` creates an `sqlalchemy` engine and leaves it hanging after its done. I managed to reproduce it many time on an M1 macbook pro and fix it by changing the code to dispose the engine right before `_connect` returns.

Here is what the warning looks like:
```
../../../Library/Caches/pypoetry/virtualenvs/python-testkit-5lwKVr0p-py3.10/lib/python3.10/site-packages/_pytest/runner.py:341:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../../../Library/Caches/pypoetry/virtualenvs/python-testkit-5lwKVr0p-py3.10/lib/python3.10/site-packages/_pytest/runner.py:262: in <lambda>
    lambda: ihook(item=item, **kwds), when=when, reraise=reraise
../../../Library/Caches/pypoetry/virtualenvs/python-testkit-5lwKVr0p-py3.10/lib/python3.10/site-packages/pluggy/_hooks.py:433: in __call__
    return self._hookexec(self.name, self._hookimpls, kwargs, firstresult)
../../../Library/Caches/pypoetry/virtualenvs/python-testkit-5lwKVr0p-py3.10/lib/python3.10/site-packages/pluggy/_manager.py:112: in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
../../../Library/Caches/pypoetry/virtualenvs/python-testkit-5lwKVr0p-py3.10/lib/python3.10/site-packages/_pytest/unraisableexception.py:88: in pytest_runtest_call
    yield from unraisable_exception_runtest_hook()
[...]

PR #380, opened 2023-08-17
fix: Close docker client when stopping the docker container

rstcruzo:
Fixes the following warning

```
sys:1: ResourceWarning: unclosed <socket.socket fd=7, family=AddressFamily.AF_UNIX, type=SocketKind.SOCK_STREAM, proto=0, raddr=/Users/rodrigo/.docker/run/docker.sock>
```
 Related to #379


alexanderankin:
can you add a reproducer? i cannot replicate with:
<details><summary>PostgresContainer.py</summary>

```python
from typing import Any

from sqlalchemy import create_engine, Engine, Connection, CursorResult, text

p = PostgresContainer("postgres:14-alpine")


def main():
    p.start()
    engine: Engine = create_engine(p.get_connection_url())
    with engine.connect() as conn:
        conn: Connection = conn
        conn.execute(text("create table if not exists hero(name varchar(500) unique)"))
        conn.execute(text("insert into hero(name) values ('alice')"))
        conn.execute(text("insert into hero(name) values ('bob')"))
        with conn.execute(text('select * from hero')) as execute:
            execute: CursorResult[Any] = execute
            print(execute.fetchall())


if __name__ == '__main__':
    main()
```

</details>

PR #388, opened 2023-10-20
feat(core)!: add support for `tc.host` and de-prioritise `docker:dind`

kiview:
## What does this PR do?

Adds support for reading the `tc.host` property from the `~/.testcontainers.properties` file. This config will have the highest priority for configuring the Docker client.

## Why is it important

This brings [Testcontainers Desktop](https://testcontainers.com/desktop/) support to testcontainers-python and aligns this language with the `TestcontainersHostStrategy` found it testcontainers-java.

## Misc

I wasn't able to get a working testcontainers-python development environment set up on my M2 MacBook. This seems to be related to some kind of Cython 3.0.0 issue, that I don't fully understand and trying to fix it breaks the installation of further dependencies (like pymssql). So I was only able to test it in an Ubuntu Codespaces environment (also required some weird Cython workarounds).

##  Breaking Changes

- To prioritise `tc.host` this PR prioritised having that over true `docker:dind` use cases
- This means we stopped trying to automatically infer the container host IP when running inside a `docker:dind` container
- When using `-v /var/run/docker.sock:/var/run/docker.sock` you should be unaffected since your containers run on the original host


core/testcontainers/core/container.py
```diff
@@ -90,18 +90,18 @@ def get_container_host_ip(self) -> str:
...
-            # container IP address, otherwise fall back to detected host
-            # address. Even it's inside container, we need to double check,
-            # because docker host might be set to docker:dind, usually in CI/CD environment
-            gateway_ip = self.get_docker_client().gateway_ip(self._container.id)
-
-            if gateway_ip == host:
-                return self.get_docker_client().bridge_ip(self._container.id)
-            return gateway_ip
+        # # check testcontainers itself runs inside docker container
+        # if inside_container() and not os.getenv("DOCKER_HOST") and not host.startswith("http://"):
```
kiview, line 94: Had to do it, to get things working with Testcontainers Cloud within Codespaces (which runs inside a container). I also thought part of this code duplicate what is done in `docker_client.py`.

core/testcontainers/core/docker_client.py
```diff
@@ -23,15 +25,8 @@
...
 LOGGER = setup_logger(__name__)
-
-
-def _stop_container(container: Container) -> None:
-    try:
-        container.stop()
-    except NotFound:
-        pass
-    except Exception as ex:
-        LOGGER.warning("failed to shut down container %s with image %s: %s", container.id, container.image, ex)
```
totallyzen, line 34: private function -> moved to the bottom of this file

totallyzen (approved):
After some cleanup of the "read file" mechanism, it's LGTM.
Also updated the PR body to add details on why this is potentially a breaking change and what to do about it if you are affected.

massdosage:
We just got bitten by this change when we upgraded from version `3.5.0` to `4.8.1`. We have a unit test which spins up a `DockerContainer` which has an application that exposes a REST end point over HTTP. We connect to this from the unit test to check it's available. So we used to do something like this:

```
        with DockerContainer(container).with_exposed_ports(8083) as container:
            url = f'http://{container.get_container_host_ip()}:{container.get_exposed_port(8083)}/'
            requests.get(url).raise_for_status()
```
When we run this in our CI pipeline it is using Docker within Docker and the URL generated above is `http://172.17.0.1:32769/` for version `3.5.0`. When we upgraded to `4.8.1` the URL changes to be of the form `http://localhost:32769` which doesn't resolve and the test fails with a `ConnectionRefusedError`. I tried various forms of setting a value for `tc.host` such as `tc.host=http://172.17.0.1:32769`, `tc.host=tcp://172.17.0.1:1234` and they all resulted in connection errors.

Eventually I resorted to moving the old behaviour that I found on this PR into my unit test like so

```
            if inside_container():
                host = container.get_docker_client().gateway_ip(container.get_wrapped_container().id)
            else:
                host = container.get_container_host_ip()
            url = f'http://{host}:{container.get_exposed_port(8083)}/'
```
and this works.

Am I doing something completely wrong or is this expected behaviour?

kiview:
Ah sorry @massdosage ,I see the community (@CarliJoy) is already working on it, maybe you can try to stay in the loop there:
https://github.com/testcontainers/testcontainers-python/issues/517#issuecomment-2348478653

massdosage:
@kiview no problem, I know how these things go. I've joined the discussion on #517 and would be happy to review or try out any proposed changed.

Issue #391, opened 2023-11-08
flaky tests


alexanderankin:
**Describe the bug**

tests flaky locally

**To Reproduce**

run tests multiple times locally

**Runtime environment**

standard ubuntu

---

creating issue to link several prs together

alexanderankin:
this is not really so necessary as the majority of blocking level concerns have been addressed - i think a review of the waiting strategies is not going to be forgotten - its a theme in the issues - lets just do things the way other tc libs implement them - an audit is necessary but this issue staying open doesn't help it

alexanderankin:
one of the other flaky ones was elastic - to set the memory, we should implement transferable to write the config file (rather than mount) - again, kind of questionable to do it under the banner of this ticket, but reminding myself so it doesn't get lost

PR #389, opened 2023-10-29
feat(new): add a new Docker Registry test container

max-pfeiffer:
I added a new test container for spinning up a [Docker registry](https://hub.docker.com/_/registry).

PR #392, opened 2023-11-10
#391 build test fixes

alexanderankin:
this pull request updates the code so that tests are green when code is correct. it was flaky before, and some code outdated, giving false negative result

* nginx - fix implementation - add waiting
* give elastic more memory so it is more likely to succeed
* selenium - update argument name (their constructor api changed)
* formatting
* comment out pyyaml dep in docker-compose module (also comment all tests)

opened re: #391

### DIFF
--- a/compose/setup.py
+++ b/compose/setup.py
@@ -12,7 +12,7 @@ setup(
     url="https://github.com/testcontainers/testcontainers-python",
     install_requires=[
         "testcontainers-core",
-        "docker-compose",
+        # "docker-compose",
     ],
     python_requires=">=3.7",
 )
--- a/compose/setup.py
+++ b/compose/setup.py
@@ -12,6 +12,7 @@ setup(
     url="https://github.com/testcontainers/testcontainers-python",
     install_requires=[
         "testcontainers-core",
+        # docker compose support has been removed to fix dependency issues in the build
         # "docker-compose",
     ],
     python_requires=">=3.7",
--- a/compose/tests/test_docker_compose.py
+++ b/compose/tests/test_docker_compose.py
@@ -1,3 +1,4 @@
+"""
 import os
 from unittest.mock import patch
 
@@ -9,8 +10,14 @@ from testcontainers.core.exceptions import NoSuchPortExposed
 from testcontainers.core.waiting_utils import wait_for_logs
 
 ROOT = os.path.dirname(__file__)
+"""
 
 
+def test_can_spawn_service_via_compose():
+    assert 2 == 2
+
+
+"""
 def test_can_spawn_service_via_compose():
     with DockerCompose(ROOT) as compose:
         host = compose.get_service_host("hub", 4444)
@@ -123,3 +130,4 @@ def test_can_exec_commands():
         assert result[0] == 'my_test\n', "The echo should be successful"
         assert result[1] == '', "stderr should be empty"
         assert result[2] == 0, 'The exit code should be successful'
+"""
--- a/compose/tests/test_docker_compose.py
+++ b/compose/tests/test_docker_compose.py
@@ -11,7 +11,7 @@ from testcontainers.core.waiting_utils import wait_for_logs
 ROOT = os.path.dirname(__file__)
 
 
-@pytest.mark.skip(reason="docker compose support has been removed to fix dependency issues in the build")
+@pytest.mark.skip(reason="compose support has been removed to fix dependency issues in the build")
 def test_can_spawn_service_via_compose():
     with DockerCompose(ROOT) as compose:
         host = compose.get_service_host("hub", 4444)
@@ -20,7 +20,7 @@ def test_can_spawn_service_via_compose():
         assert port == "4444"
 
 
-@pytest.mark.skip(reason="docker compose support has been removed to fix dependency issues in the build")
+@pytest.mark.skip(reason="compose support has been removed to fix dependency issues in the build")
 def test_can_pull_images_before_spawning_service_via_compose():
     with DockerCompose(ROOT, pull=True) as compose:
         host = compose.get_service_host("hub", 4444)
@@ -29,7 +29,7 @@ def test_can_pull_images_before_spawning_service_via_compose():
         assert port == "4444"
 
 
-@pytest.mark.skip(reason="docker compose support has been removed to fix dependency issues in the build")
+@pytest.mark.skip(reason="compose support has been removed to fix dependency issues in the build")
 def test_can_build_images_before_spawning_service_via_compose():
     with patch.object(DockerCompose, "_call_command") as call_mock:
         with DockerCompose(ROOT, build=True) as compose:
@@ -42,7 +42,7 @@ def test_can_build_images_before_spawning_service_via_compose():
     assert "--build" in docker_compose_cmd
 
 
-@pytest.mark.skip(reason="docker compose support has been removed to fix dependency issues in the build")
+@pytest.mark.skip(reason="compose support has been removed to fix dependency issues in the build")
 def test_can_specify_services():
     with patch.object(DockerCompose, "_call_command") as call_mock:
         with DockerCompose(ROOT, services=["hub", "firefox"]) as compose:
@@ -56,7 +56,7 @@ def test_can_specify_services():
     assert "chrome" not in docker_compose_cmd
 
 
-@pytest.mark.skip(reason="docker compose support has been removed to fix dependency issues in the build")
+@pytest.mark.skip(reason="compose support has been removed to fix dependency issues in the build")
 @pytest.mark.parametrize("should_run_hub", [
     [True],
     [False],
@@ -76,27 +76,27 @@ def test_can_run_specific_services(should_run_hub: bool):
                 assert compose.get_service_host("hub", 4444)
 
 
-@pytest.mark.skip(reason="docker compose support has been removed to fix dependency issues in the build")
+@pytest.mark.skip(reason="compose support has been removed to fix dependency issues in the build")
 def test_can_throw_exception_if_no_port_exposed():
     with DockerCompose(ROOT) as compose:
         with pytest.raises(NoSuchPortExposed):
             compose.get_service_host("hub", 5555)
 
 
-@pytest.mark.skip(reason="docker compose support has been removed to fix dependency issues in the build")
+@pytest.mark.skip(reason="compose support has been removed to fix dependency issues in the build")
 def test_compose_wait_for_container_ready():
     with DockerCompose(ROOT) as compose:
         docker = DockerClient()
         compose.wait_for("http://%s:4444/wd/hub" % docker.host())
 
 
-@pytest.mark.skip(reason="docker compose support has been removed to fix dependency issues in the build")
+@pytest.mark.skip(reason="compose support has been removed to fix dependency issues in the build")
 def test_compose_can_wait_for_logs():
     with DockerCompose(filepath=ROOT, compose_file_name="docker-compose-4.yml") as compose:
         wait_for_logs(compose, "Hello from Docker!")
 
 
-@pytest.mark.skip(reason="docker compose support has been removed to fix dependency issues in the build")
+@pytest.mark.skip(reason="compose support has been removed to fix dependency issues in the build")
 def test_can_parse_multiple_compose_files():
     with DockerCompose(filepath=ROOT,
                        compose_file_name=["docker-compose.yml", "docker-compose-2.yml"]) as compose:
@@ -111,7 +111,7 @@ def test_can_parse_multiple_compose_files():
         assert port == "4444"
 
 
-@pytest.mark.skip(reason="docker compose support has been removed to fix dependency issues in the build")
+@pytest.mark.skip(reason="compose support has been removed to fix dependency issues in the build")
 def test_can_get_logs():
     with DockerCompose(ROOT) as compose:
         docker = DockerClient()
@@ -120,7 +120,7 @@ def test_can_get_logs():
         assert stdout, 'There should be something on stdout'
 
 
-@pytest.mark.skip(reason="docker compose support has been removed to fix dependency issues in the build")
+@pytest.mark.skip(reason="compose support has been removed to fix dependency issues in the build")
 def test_can_pass_env_params_by_env_file():
     with DockerCompose(ROOT, compose_file_name='docker-compose-3.yml',
                        env_file='.env.test') as compose:
@@ -128,7 +128,7 @@ def test_can_pass_env_params_by_env_file():
         assert stdout.splitlines()[0], 'test_has_passed'
 
 
-@pytest.mark.skip(reason="docker compose support has been removed to fix dependency issues in the build")
+@pytest.mark.skip(reason="compose support has been removed to fix dependency issues in the build")
 def test_can_exec_commands():
     with DockerCompose(ROOT) as compose:
         result = compose.exec_in_container('hub', ['echo', 'my_test'])
--- a/compose/tests/test_docker_compose.py
+++ b/compose/tests/test_docker_compose.py
@@ -1,4 +1,3 @@
-"""
 import os
 from unittest.mock import patch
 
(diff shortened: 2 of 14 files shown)


alexanderankin:
btw, not sure if we want to proactively handle this type of thing with regards to ipv6 ips:

```diff
diff --git a/nginx/testcontainers/nginx/__init__.py b/nginx/testcontainers/nginx/__init__.py
index d0680f1..c4bcbf7 100644
--- a/nginx/testcontainers/nginx/__init__.py
+++ b/nginx/testcontainers/nginx/__init__.py
@@ -37,5 +37,14 @@ class NginxContainer(DockerContainer):
 
     @wait_container_is_ready(urllib.error.URLError)
     def _connect(self, host: str, port: str) -> None:
-        url = urllib.parse.urlunsplit(('http', f'{host}:{port}', '', '', ''))
+        url = urllib.parse.urlunsplit(('http', self._netloc_of(host, port), '', '', ''))
         urllib.request.urlopen(url, timeout=1)
+
+    @staticmethod
+    def _netloc_of(host: str, port: str):
+        import ipaddress
+        try:
+            version = ipaddress.ip_address(host).version
+            return f"{host if version == 4 else f'[{host}]'}:{port}"
+        except ValueError:  # when "localhost" instead of ip
+            return f'{host}:{port}'
```

alexanderankin:
one other potential failure is the lack of python-dev package as its necessary to compile some c code for oracle db driver, but i think this is easy enough to add once we see that this is why it fails.

kiview:
We have dabbled with the IPv4 & IPv6 topic in the other Testcontainers languages, and I think we have a robust approach for this. But let's move it out of the scope of this PR 👍

kiview:
Great, all tests are green now 👍 

@alexanderankin could you please add a description to the PR detailing some of the changes and the rationale behind it, for future documentation?

compose/tests/test_docker_compose.py
```diff
@@ -9,8 +10,14 @@
...
 
 ROOT = os.path.dirname(__file__)
+"""
 
 
+def test_can_spawn_service_via_compose():
+    assert 2 == 2
+
+
+"""
```
kiview, line 20: Can we use a pytest feature to skip these tests?

```suggestion
"""
@pytest.mark.skip(reason="docker compose support has been removed to fix dependency issues in the build")
"""
```

compose/setup.py
```diff
@@ -12,7 +12,7 @@
     url="https://github.com/testcontainers/testcontainers-python",
     install_requires=[
         "testcontainers-core",
-        "docker-compose",
+        # "docker-compose",
```
kiview, line 15: Let's add a comment explaining why this is now commented.

compose/tests/test_docker_compose.py
```diff
@@ -123,3 +130,4 @@ def test_can_exec_commands():
         assert result[0] == 'my_test\n', "The echo should be successful"
         assert result[1] == '', "stderr should be empty"
         assert result[2] == 0, 'The exit code should be successful'
+"""
```
kiview, line 133:
```suggestion
```

Not needed to comment if we skip (I hope).

elasticsearch/testcontainers/elasticsearch/__init__.py
```diff
@@ -88,7 +88,7 @@ def __init__(self, image: str = "elasticsearch", port: int = 9200, **kwargs) ->
 
     @wait_container_is_ready(URLError)
     def _connect(self) -> None:
-        res = urllib.request.urlopen(self.get_url())
+        res = urllib.request.urlopen(self.get_url(), timeout=1)
```
kiview, line 91: What is the default timeout if we don't add this? Infinity?

elasticsearch/tests/test_elasticsearch.py
```diff
@@ -1,5 +1,5 @@
 import json
-import urllib
+import urllib.request
```
kiview, line 2: Since there are no other changes in this file, is this necessary?

elasticsearch/testcontainers/elasticsearch/__init__.py
```diff
@@ -12,7 +12,7 @@
 #    under the License.
 import logging
 import re
-import urllib
+import urllib.request
```
kiview, line 15: Why specify it like this, if we use it fully qualified below?

elasticsearch/tests/test_elasticsearch.py
```diff
@@ -8,6 +8,8 @@
 # The versions below were the current supported versions at time of writing (2022-08-11)
 @pytest.mark.parametrize('version', ['6.8.23', '7.17.5', '8.3.3'])
 def test_docker_run_elasticsearch(version):
-    with ElasticSearchContainer(f'elasticsearch:{version}') as es:
+    # adding the mem_limit as per: https://stackoverflow.com/a/75701019
+    # could also add (but not necessary): .with_env('discovery.type', 'single-node')
+    with ElasticSearchContainer(f'elasticsearch:{version}', mem_limit='3G') as es:
```
kiview, line 13: In tc-java, we configure the Elastic process using 
```
-Xms2147483648
-Xmx2147483648
```
don't we need to do the same here?

nginx/testcontainers/nginx/__init__.py
```diff
@@ -20,3 +25,17 @@ def __init__(self, image: str = "nginx:latest", port: int = 80, **kwargs) -> Non
         super(NginxContainer, self).__init__(image, **kwargs)
         self.port = port
         self.with_exposed_ports(self.port)
+
+    def start(self) -> 'NginxContainer':
```
kiview, line 29: Were there flaky tests without this change?

selenium/testcontainers/selenium/__init__.py
```diff
@@ -24,7 +25,7 @@
 }
 
 
-def get_image_name(capabilities: str) -> str:
+def get_image_name(capabilities: Dict[str, Any]) -> str:
```
kiview, line 28: Isn't this a breaking change in the API?

selenium/tests/test_selenium.py
```diff
@@ -4,7 +4,9 @@
 from testcontainers.core.utils import is_arm
 
 
-@pytest.mark.parametrize("caps", [DesiredCapabilities.CHROME, DesiredCapabilities.FIREFOX])
+@pytest.mark.parametrize("caps",
+                         [DesiredCapabilities.CHROME, DesiredCapabilities.FIREFOX],
+                         ids=['chrome', 'ff'])
```
kiview, line 9: I think this highlights, that above was a breaking API change. Can we do anything about this?

compose/tests/test_docker_compose.py
alexanderankin: the danger with that is that empty test should be reported as a failure (imo its a bug if pytest passes with 0 tests) - next/soon pr should be removal of docker-compose altogether, anyway right?

elasticsearch/testcontainers/elasticsearch/__init__.py
alexanderankin: i think it is infinity, yes

in the future when we add strategies and all the oop fun/gunk (i havent worked on tc-java enough to know which it is) for default waiting etc, then we can centralize this. but for now, want to proliferate it so it is not lost.

elasticsearch/tests/test_elasticsearch.py
alexanderankin: i believe so- it is relying on some other module importing the correct submodule:

```console
toor@z14-1:~$ python
Python 3.11.6 (main, Oct 23 2023, 22:48:54) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import urllib
>>> urllib.request.urlopen
  File "<stdin>", line 1, in <module>
AttributeError: module 'urllib' has no attribute 'request'
>>> 
```

elasticsearch/tests/test_elasticsearch.py
alexanderankin: hm, i can update this to 2g to make it match

selenium/testcontainers/selenium/__init__.py
alexanderankin: it is, but the api definition was incorrect before - it is indexed in the old code like a dict anyways to determine which image to use (ff vs chrome)

selenium/testcontainers/selenium/__init__.py
```diff
@@ -24,7 +25,7 @@
...
     return IMAGES[capabilities['browserName']]
```
alexanderankin, line 30: here is the full method

selenium/tests/test_selenium.py
alexanderankin: all im doing here is naming the test cases, so you can see which one fails, if one of them does fail, purely cosmetic

selenium/testcontainers/selenium/__init__.py
alexanderankin: resolved this, copied to new discussion with whole method quoted

elasticsearch/tests/test_elasticsearch.py
alexanderankin: oh, so - it depends on how much faith we have in the commitment to backwards compat on the side of the elastic team

in the logs, the container prints a 1.5g heap on startup. do we think that will change? otherwise, i guess i have to explicitly set the heap, so that we know the mem_limit will be high enough in the future

compose/tests/test_docker_compose.py
alexanderankin: looking into this locally

alexanderankin: ok, pushed fix for this one

compose/setup.py
alexanderankin: added commit for this:

commit 2bd20c6a9ab811343beee150a4816711848fb9ea
Author: Dave Ankin <daveankin@gmail.com>
Date:   Mon Nov 13 06:18:08 2023 -0500

    #392: adding comment to explain why testcontainers-compose no longer depends on docker-compose

diff --git a/compose/setup.py b/compose/setup.py
index 9b783f0..a0b0f84 100644
```diff
--- a/compose/setup.py
+++ b/compose/setup.py
@@ -12,6 +12,7 @@ setup(
     url="https://github.com/testcontainers/testcontainers-python",
     install_requires=[
         "testcontainers-core",
+        # docker compose support has been removed to fix dependency issues in the build
         # "docker-compose",
     ],
     python_requires=">=3.7",
```

compose/tests/test_docker_compose.py
alexanderankin: resolving this to continue discussion where the pytest.mark.skip is suggested

elasticsearch/tests/test_elasticsearch.py
alexanderankin: should i implement `with_copy_file_to_container` in this same pr? it is missing for parity with tc-java

nginx/testcontainers/nginx/__init__.py
alexanderankin: yes

elasticsearch/testcontainers/elasticsearch/__init__.py
kiview: It means without adding this, you could see behavior where we would wait indefinitely for the connection (maybe because there was an issue during container startup) and so the test suite would hang?

elasticsearch/tests/test_elasticsearch.py
kiview: No, let's not add features in this PR. My comment above was about tc-java configuring the ES JVM specifically to use 2GB.

I am pretty sure the used heap will else depend on the available memory.

What is the intent behind the 3G limit here? To not have the ES container use up all memory of the Docker host?

selenium/testcontainers/selenium/__init__.py
kiview: Does this mean using the old API with a `str` would have lead to an error? So we don't change the API (it was always a `dict`), we just fix the type hint?

elasticsearch/tests/test_elasticsearch.py
alexanderankin: without the 3G specified, it failed to start on my machines (ubuntu 16/32 gig ram). sure, there are some fancy ways containers/apps/jvm determine their memory sizes, but from my experience, it just helps to be explicit.

requirements/macos-latest-3.10.txt
```diff
@@ -138,8 +138,6 @@ docker[ssh]==6.1.3
     # via
     #   docker-compose
     #   testcontainers-core
-docker-compose==1.29.2
```
kiview, line 141: I don't think these files should be manually edited, since they contain the following comment:
```
# This file is autogenerated by pip-compile with Python 3.10
# by the following command:
#
#    pip-compile --output-file=requirements.txt --resolver=backtracking
```

kiview:
Unfortunately, the build still fails on MacOS because of issues installing `pymssql`, but we can fix this in another PR.

requirements/macos-latest-3.10.txt
alexanderankin: That is true but I cannot auto generate this with the required os/env.

elasticsearch/testcontainers/elasticsearch/__init__.py
alexanderankin: Yeah this works because another python module does the full import, so it's dependent on global state probably not what you want

alexanderankin: I posted a reproducer above/below

max-pfeiffer:
@kiview Because of the issue with `pymssql` on MacOS: I think I can help here. 

It's related to the feetds build dependency for `pymssql` package. You need to have freetds installed on your machine to be able to build the `pymssql` package.

I am using [MacPorts](https://www.macports.org/). This way it works for me:
```
sudo port install freetds
export CFLAGS="-I/opt/local/include/freetds"
export LDFLAGS="-L/opt/local/lib/"
pip install pymssql==2.2.10
```
If you use Homebrew you need to point the flags to your library directories.

I just pulled the lastest main branch and I can install the dependencies. Thanks a lot for your help here. Much appreciated.

alexanderankin:
it seems like a strong argument to drop the dependency or make it optional

max-pfeiffer:
Well, isn't it already optional when using the package? You could just install the features you need: `pip install testcontainers-<feature>`

As a software engineer I do not mind tackling with these kind of problems when working on the package and/or building it. It's a common problem on MacOs that sometimes dependencies are a bit cumbersome to install and configure.

I would advocate to add a wiki section to the project as a go-to place for such problems. Other projects like pyenv do it like this: [https://github.com/pyenv/pyenv/wiki/Common-build-problems](https://github.com/pyenv/pyenv/wiki/Common-build-problems)
I find that very helpful. I could provide the initial content if you guys agree to this approach.

We still need to fix the tests on MacOs. I noticed a bunch of tests failing when working my pull request https://github.com/testcontainers/testcontainers-python/pull/389. This is due to older image versions not supporting ARM CPUs.

Is there any chance we can get https://github.com/testcontainers/testcontainers-python/pull/357 merged? I would also volunteer to work on it if help is needed there.

alexanderankin:
i think there are some neat things in this PR we should merge but its no longer productive to keep this open in its current form: closing.

elasticsearch/tests/test_elasticsearch.py
alexanderankin: the import state is global - so if another module fully imports urllib (and chances are high enough i guess, or this happens inside requests module, or something...) then it works. but its only correct by accident:

anyways:

```
$ python -c 'import this' | grep implicit
Explicit is better than implicit.
```

elasticsearch/tests/test_elasticsearch.py
alexanderankin: this is one of the outstanding items of #391 but it makes most sense to merge after implementing transferable

alexanderankin:
ive also very recently gotten a mac so "excited" ( :slightly_smiling_face: ) to get familiar with those issues and if i can, help/start to document

PR #395, opened 2023-11-22
Add wait strategy behavior to `NginxContainer`

alexanderankin:
Adds a http-like wait strategy behavior to `NginxContainer`, to fix race condition in the tests and with general container usage.

Relates to #391.


nginx/testcontainers/nginx/__init__.py
```diff
@@ -20,3 +25,17 @@ def __init__(self, image: str = "nginx:latest", port: int = 80, **kwargs) -> Non
...
+        host = self.get_container_host_ip()
+        port = str(self.get_exposed_port(self.port))
+        self._connect(host, port)
+
+        return self
+
+    @wait_container_is_ready(urllib.error.URLError)
+    def _connect(self, host: str, port: str) -> None:
+        url = urllib.parse.urlunsplit(('http', f'{host}:{port}', '', '', ''))
+        urllib.request.urlopen(url, timeout=1)
```
kiview, line 41:
```suggestion
    def start(self) -> 'NginxContainer':
        super().start()
        self._connect()

        return self

    @wait_container_is_ready(urllib.error.URLError)
    def _connect(self) -> None:
        host = self.get_container_host_ip()
        port = str(self.get_exposed_port(self.port))
        url = urllib.parse.urlunsplit(('http', f'{host}:{port}', '', '', ''))
        urllib.request.urlopen(url, timeout=1)
```

WDYT? Just thought it makes it a bit easier to read (and it's how the `rabbitmq` module roughly structures it).

kiview (approved):

nginx/testcontainers/nginx/__init__.py
alexanderankin: this changes the structure as:

```
        host = self.get_container_host_ip()
        port = str(self.get_exposed_port(self.port))
```

are wrapped in their own decorators.

alexanderankin: the goal of extracting those calls outside the method is to minimize the printed log messages (which appear as log messages "waiting for container" or something like this)

kiview: Ok, I see. I think the calls in `start()` should not got through a method decorated with `@wait_container_is_ready()`, but that is something we should fix through a change in the design in a follow-up. 

`RabbitMqContainer` follows a similar approach here:
https://github.com/alexanderankin/testcontainers-python/blob/d02dff9f2b909dce96c1d382b03fd0ecdf9f38c2/rabbitmq/testcontainers/rabbitmq/__init__.py#L49

PR #396, opened 2023-11-22
Add support for Codespaces

kiview:
## What does this PR do?

Adds pre-configured support for GitHub Codespaces and local development using dev containers.

## Why is it important?

Makes maintenance, contributions and reviewing PRs easier, since it does not required developers to maintain a permanent local development environment (which is a bit more challenging, given the current approach of managing dependencies).


eddumelendez (approved):
I already tried and it works fine.

PR #408, opened 2024-01-02
feat(build): use poetry and organise modules

totallyzen:
> [!WARNING]
> This is a Disruptive PR, but for good reasons. We will engage with people on existing PRs and Issues to help them transition.

# Changes

Fixes #204 
- We do not publish `testcontainers-*` packages any more as it's a maintenance nightmare on PyPI 
  - example: in case of loss of access to the PyPI package, it's a messy and long procedure to reclaim it
- Instead the project root is now a collection of modules, see `pyproject.toml` 
- All published under `testcontainers` which will allow the previous `testcontainers[extra1, extra2]` format to be used (replacing the `/meta` package mechanism)
- Removes old config files like `.coveragerc` and `setup.cfg` and integrates them into `pyproject.toml` - more grooming incoming in future PRs (like the move to `ruff` which is a more performant formatter/linter)

# TODO
- [x] groom `Makefile` to accomodate the new layout
- [x] accommodate the new layout in CI
- [ ] edit repository settings to simplify the list of required checks (which matches the new layout and CI) - this is done via a GitHub App used by the `testcontainers` org

PR #411, opened 2024-01-15
Disable workflows and add issue templates

alexanderankin:
temporary changes to reduce mental overhead of dealing with automation


totallyzen (approved):
LGTM - we'll rebuild from scratch with regards to the `CODEOWNERS`

totallyzen:
looking forward to reducing the 100+ checks to a dozen or two in #408 :joy:

PR #419, opened 2024-02-09
Create settings.yaml

alexanderankin:


.github/settings.yaml
```diff
@@ -0,0 +1,156 @@
...
+    # Branch Protection settings. Set to null to disable
+    protection:
+      # Required. Require at least one approving review on a pull request, before merging. Set to null to disable.
+      required_pull_request_reviews:
+        # The number of approvals required. (1-6)
+        required_approving_review_count: 1
+        # Dismiss approved reviews automatically when a new commit is pushed.
+        dismiss_stale_reviews: false
+        # Blocks merge until code owners have reviewed.
+        require_code_owner_reviews: false
```
balint-backmaker, line 136: I think we do need to make this `true` and add a `.github/CODEOWNERS` with this at the least:

```
.github/CODEOWNERS @testcontainers/oss-matinainers
.github/settings.yaml @testcontainers/oss-matinainers
```

Why we need this?
- the app that is running this sync process has an explicit security warning about this
- their suggested workaround is the above mentioned requirement plus the `CODEOWNERS` change

balint-backmaker (changes requested):
one thing that I think we need to comply with - @kiview I think you'd agree but please lemme know

balint-backmaker (approved):
having seen your other message on slack, LGTM

totallyzen (approved):
having seen your other message on slack, LGTM

PR #420, opened 2024-02-10
Rename settings.yaml to settings.yml

alexanderankin:


totallyzen (approved):
LGTM - I think you are right because `testcontainters-go` [does the same thing](https://github.com/testcontainers/testcontainers-go/blob/main/.github/settings.yml)

PR #421, opened 2024-02-12
try adding a required check to see if it helps

alexanderankin:


totallyzen (approved):
LGTM, let's hope it takes effect!

alexanderankin:
Did it work? Should we try with pull_request in the name?

PR #422, opened 2024-02-13
fix: changed files breaks on main

totallyzen:
# change

Fixes an issue with the usage of https://github.com/tj-actions/changed-files.
Without a checkout, it breaks on `main`. This re-introduces that push to main.


totallyzen:
For the record, it does not show any checks, because we haven't changed any module files.

PR #423, opened 2024-02-13
fix: flaky garbage collection resulting in testing errors

### LINKED ISSUE #399: 'DockerContainer' object has no attribute '_container'
**Creating Postgres Container instance results in error: 'PostgresContainer' object has no attribute '_container'**

I tried starting a Postgres container and got the following error:
```
Exception ignored in: <function DockerContainer.__del__ at 0x7f1da5122b80>
  File "/.venv/lib/python3.9/site-packages/testcontainers/core/container.py", line 80, in __del__
AttributeError: 'PostgresContainer' object has no attribute '_container'
```
I already saw an issue [#353](https://github.com/testcontainers/testcontainers-python/issues/353) which was about the same error but it seems to persist.
See the code I used below.

**To Reproduce**

```python

def get_postgres_uri():
    # Use a PostgreSQL Docker container for testing
    with PostgresContainer("postgres:latest") as postgres:
        return postgres.get_connection_url()
```
I also tried out the code in the [postgres/tests/test_postgres.py](https://github.com/testcontainers/testcontainers-python/blob/main/postgres/tests/test_postgres.py) file and it failed with the same error message.

**Runtime environment**

OS: Pop!_OS 22.04 LTS x86_64
Python version: Python 3.9.12
Docker version: 24.0.5
testcontainers version: 3.7.1

totallyzen:
# change

Fixes #399. Applied a bit of defensive coding and attempted to create some tests for it, however reproducing it with a local dev machine is not easy. I did my best to reproduce the issue with garbage collection in the new test.

### DIFF
--- a/core/testcontainers/core/container.py
+++ b/core/testcontainers/core/container.py
@@ -23,6 +23,7 @@ class DockerContainer:
         >>> with DockerContainer("hello-world") as container:
         ...    delay = wait_for_logs(container, "Hello from Docker!")
     """
+
     def __init__(self, image: str, docker_client_kw: Optional[dict] = None, **kwargs) -> None:
         self.env = {}
         self.ports = {}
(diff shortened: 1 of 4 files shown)


alexanderankin (approved):
lots of unrelated changes but checking for none makes sense.

totallyzen:
> lots of unrelated changes but checking for none makes sense.

sorry, I'll be cleaner next time - I started on the `mypy` cleanup and it's a doozy!

PR #425, opened 2024-02-19
chore(lint): update typing and linting

totallyzen:
# change

advances #305
- replace `flake8` with `ruff` and install `pre-commit`
- run linters as part of a separate job (to separate linting from testing)
- add a draft of `mypy` setup but not try to fix stuff because it's a complicated procedure to fix everything - this PR already reformatted a lot of code


alexanderankin:
https://github.com/testcontainers/testcontainers-python/actions/runs/7979817068/job/21788105968?pr=425#step:8:29 - hopefully fixed in https://github.com/testcontainers/testcontainers-python/pull/428 but at any rate doesn't need to be fixed here

PR #438, opened 2024-03-02
fix(build): add `pre-commit` as a dev dependency to simplify local dev and CI

max-pfeiffer:
I could not install pre-commit handler because the package was missing in [tool.poetry.group.dev.dependencies].

I also updated the configuration for the pre-commit handler as black and ruff packages were outdated. Also removed deprecation for ruff v0.3.0 config in pyproject.toml.


pyproject.toml
```diff
@@ -108,12 +108,15 @@ redis = ["redis"]
...
 [tool.poetry.group.dev.dependencies]
+black = "24.2.0"
+mypy = "1.7.1"
+pre-commit = "3.6.2"
+pg8000 = "*"
 pytest = "7.4.3"
 pytest-cov = "4.1.0"
+ruff = "0.3.0"
 sphinx = "^7.2.6"
-pg8000 = "*"
```
alexanderankin, line 114: Pretty sure this has to be here

pyproject.toml
```diff
@@ -108,12 +108,15 @@ redis = ["redis"]
...
 twine = "^4.0.2"
-mypy = "1.7.1"
```
alexanderankin, line 116: There are plans to add this so let's keep it also

pyproject.toml
max-pfeiffer: It's still there, see above. I just sorted this dependency section alphabetically.

pyproject.toml
max-pfeiffer: It's still there, see above. I just sorted this dependency section alphabetically.

alexanderankin:
i think #440 is preferable as it does not introduce any extraneous changes, @totallyzen feel free to pick one and merge.

additionally we have an incorrect ruff config it looks like, which is fixed here, which i would rather 1) fix in a separate PR 2) fix after some releases because it may change the linting which just introduces more work (im not a big linting believer but we should *eventually* fix our config i guess)

pyproject.toml
alexanderankin: whoops, my bad - totally missed that.

totallyzen:
@max-pfeiffer 
There is an argument _against_ installing `pre-commit` as a dev dependency.
It's because it's another package manager that then hooks into git. This way "who is on top?" becomes a problem via poetry-ception.

However!
- I did some digging with `pre-commit` and based on their issues, they aren't against your approach
- I'm willing to recognise my own biases in this and my workflow wasn't really affected
- I did a test with your branch, it worked for me too, so despite the differences, we can argue that your setup is simpler.
- Therefore I'm happy to proceed like this, with one exception that we make it more permissive than just the strict version you've installed it with -> `^3.6` (I see you `renovate` user :wink: we'll get there on this project too, just give it time)

@alexanderankin 

> additionally we have an incorrect ruff config it looks like

It's not an incorrect one, the api "broke" between between `0.2` and `0.3` (config was made more explicit and old style was deprecated) and Max updated the config to match the new recommendation.
It's a common notion that `0.X` versions allow unstable API :pray: 

I copy-pasted some old settings from another project when I set up `ruff` for this project so I see no reason why not jump to latest practices. It is also a cleaner definition in the `pyoproject.toml`

PR #439, opened 2024-03-02
feat(nats): Add new module

bearrito:
Adds support for NATS. https://nats.io/

I wouldn't merge this yet. It's a bit awkward in determining if the container is ready (because of async) in the `start` method. 
I'll sort out how I want to do that.


bearrito:
@totallyzen Is this good to go?

totallyzen:
Hey @bearrito can you give a bit more context on the use-case this covers?
There is a _lot_ to unpack without some explanation in the `README.rst`

Cons:
- adding a new class and a new dependency for 3 lines of code
- the extra package causes `poetry lock` to become slower and slower

Alternatives:
- we now support compose v2 and intend to release it with `tc-python` 4.0!
- in `compose` you can define a healthcheck for your service and it trivialises a lot of use-cases to `docker compose up --wait` type operations

WDYT?

bearrito:
I'm not quite sure I understand the comments. This adds a new module for the Nats streaming pub/sub project.

Is the suggestion that instead of creating a new module for every container that we encapsulate that into  docker compose files?

alexanderankin:
i am also confused by the lot to unpack and the compose suggestion.

certainly you will be up and running faster with a compose file or adding this class to your project's test code than waiting for us because you can do that right away, but i dont think that means we are not adding any future modules

> If that's the case how do we add convenience methods for users without adding additional libraries to poetry?

this is the real pickle - we basically want to avoid adding other libraries public apis to our public api

i would actually offer different feedback which is that we should probably use a log waiting strategy - https://github.com/testcontainers/testcontainers-java/issues/1733

alexanderankin:
I tried to write down my priorities in one of these tickets but lost it, a
key one is pairing down dependencies - some folks unhappy with postgres
something or other - but generally waiting for logs seems to make more
people happier

bearrito:
@alexanderankin Updated with logs waiting.

Side note it might be nice to expose the max wait time for the log wait. Right now it's set in the config file (which I could abuse). I think I'd generally want it less than 2 mins especially in CI where I could end up waiting NUM_TESTS * 2 mins. Either way..

totallyzen:
Hey,

Been out for my evening. David is right, while I prefer using core
libraries, it's certainly a simpler solution if you can make it with as few
extra dependencies as possible.

Cheers!

bearrito:
@totallyzen or @alexanderankin For this go around, since I've implemented the log based wait, is this good to go (after I fix the conflict)? Or do I need to go another direction?

alexanderankin:
Let me take a look I haven't gotten a chance yet even at the old one. If it
looks good I will merge, I don't see why not. Mostly time bottlenecked on
my end.

alexanderankin (approved):
lgtm

alexanderankin:
there was no need to merge - there are no conflicts - see https://github.com/alexanderankin/testcontainers-python/pull/6

PR #441, opened 2024-03-03
fix(kafka): Add redpanda testcontainer module

gudjonragnar:


pyproject.toml
```diff
@@ -82,6 +83,7 @@ cx_Oracle = { version = "*", optional = true }
 psycopg2-binary = { version = "*", optional = true }
 pika = { version = "*", optional = true }
 redis = { version = "*", optional = true }
+redpanda = { version = "*", optional = true }
```
totallyzen, line 86: The `redpanda` lib is not what you think it is. https://pypi.org/project/redpanda/ neither is https://pypi.org/project/red-panda/

You won't need either of these since this is a kafka-compliant container. See my overall review!

pyproject.toml
```diff
@@ -105,6 +107,7 @@ oracle = ["sqlalchemy", "cx_Oracle"]
 postgres = ["sqlalchemy", "psycopg2-binary"]
 rabbitmq = ["pika"]
 redis = ["redis"]
+redpanda = []
```
totallyzen, line 110: not needed, move the container into the `kafka` module instead, please

totallyzen (dismissed):
because it's a stand-in for kafka, I'd rather have you implement this in the `kafka` module instead of a completely new module. At that point it isn't about "what the container is" but what the developer experience looks like.

I have no issues with `from testcontainers.kafka import KafkaContainer, RedpandaContainer` :pray: 
In fact it would be interesting to see if you can devise a common interface from them - that would help devs choose more easily.

alexanderankin:
i opened https://github.com/gudjonragnar/testcontainers-python/pull/2 in the meantime

alexanderankin:
FYI to all who follow testcontainers-python - the kafka package on pypi is being (temporarily?) published on a different package name, please see the issue dpkp/kafka-python#2431 for more details.

Issue #336, opened 2023-04-18
Design question / is a hard dependency on SqlAlchemy really necessary?


logicbomb:
Hi, 

It seems that the `_connect` method is causing the DbContainer project to take a hard dependency on SqlAlchemy.  This is troublesome for those of us not using that package.  Is it meant to do more than verify the database container is up & running? If not, maybe there's a more neutral way of checking like using `pg_isready`?  

https://github.com/testcontainers/testcontainers-python/blob/a2bc91eb5c387d389227117b65e6398e21b5da23/testcontainers/core/generic.py#L31

thanks,
Jason

tillahoffmann:
Yes, that's a good point. We should move away from a hard dependency on sqlalchemy as you suggested. Do you have thoughts on how to integrate that change? A PR would be very welcome!

logicbomb:
i thought to use `popen`  to execute the `pg_isready` command, does that sound like a reasonable approach?

HofmeisterAn:
> i thought to use `popen` to execute the `pg_isready` command, does that sound like a reasonable approach?

Based on my experience maintaining TC for .NET, waiting for `pg_isready` to succeed is not always reliable. There were instances where the .NET Npgsql client could not establish a connection. Since I wait for specific [log messages](https://github.com/testcontainers/testcontainers-dotnet/blob/c983feb779b0115c1c6393b88cfc35b2871306eb/src/Testcontainers.PostgreSql/PostgreSqlBuilder.cs#L131-L137), I have not encountered any connection issues.

logicbomb:
I put together a WIP here: https://github.com/logicbomb/testcontainers-python/pull/1/files

You can see 2 possible implementations, one based on `pg_isready` the other checking logs as described by @HofmeisterAn.   Using `pg_isready` is much less code and seems to work, but I'm happy to go with either.

Also, would you all require a change to the way all databases are verified, or would you accept a change only for Postgres?

tillahoffmann:
You may want to have a look at the `wait_for_logs` function [here](https://github.com/testcontainers/testcontainers-python/blob/35fc2476a5823176d07952f93f8453f20c1e6392/core/testcontainers/core/waiting_utils.py#L79). That should reduce the amount of code required.

logicbomb:
ooh, `wait_for_logs` is nice. i take it you prefer that approach?

tillahoffmann:
Yes, using `wait_for_logs` could be a good option.

[...]

PR #445, opened 2024-03-06
feat(postgres): Remove SqlAlchemy dependency from postgres container

jankatins:
Updates the pg testcontainer implementation to not use (and not install) SQLAlchemy nor psycopg2.

Closes: #340
Closes: #336
Closes: #320


jankatins:
One outstanding review comment by @totallyzen:
> Make it not be in generic.py. These "db containers" assume a certain URL format, but is it worth it to have an entire 
> class just for this? (I won't hide my distaste over DbContainer and why it is in core in the first place)

So what's the alternative (in the order of my preference, best to lowest)?

- Move it to a different file?
- Make all the db container implement ´_create_connection_url()`?
- Add a different layer, like singledispatch to turn the container into an URL?

I didn't apply any changes in any of these directions yet, the current PR is just a rebase and small fix for the dependencies and newer PG versions in tests and examples.

jankatins:
`poetry run pytest -s --log-level=DEBUG modules/postgres` passes locally :-)

totallyzen:
> So what's the alternative (in the order of my preference, best to lowest)?

Hm... I'm open to suggestions! :thinking: 

Let me think out aloud...

How I view the value of these containers... focus on 
- easy configuration
- always start ready
- give a connectivity URLs instead of clients (less enforced dependencies)

I think I'd like both of the first 2 options:
- move to new module `testcontainers.databases` sounds contextually right
- make the current db containers implement an abstract class that enforces some confidence it is hard to get wrong (like forgetting to override a method)

I am however happy if you only implement just one of those. I think you see where I'm coming from!

WDYT?

jankatins:
> move to new module testcontainers.databases sounds contextually right

Does that module life under `core/testcontainers/database` or do you mean to add a complete subdirectory in the repo root? I would curerntly assume that it's `core/testcontainers/database`. That would basically be just a rename: currently it's called `generic.py`, now it would be a `databases.py`. I would currently prefer that, moving it to a extra module which is not DB specific looks trange given the current split into "modules are service specific code, core is everything which can be used by more than one module" (thats at leatst my first impression).

(I haven't understood why this split is needed, poetry build seems to "undo" it with all the includes, the only reason I can think of is that it made splitting the packages was easier but that's seems to be not needed anymore? for having the test "nearer" to the specific code?)

Another question_ do you consider these classes API or implementation detail? E.g. if I rename `DbContainer` to `SQLAlchemyBasedDBContainer`, is that a breakign change? (if it would be a breaking change, just moving them to another module would also be a breaking change, so I hope that these are implmentation details...)

jankatins:
Ok, one additional idea for the refactoring would be to just extract the sqla based `_connect()` (with the `@wait_container_is_ready(*ADDITIONAL_TRANSIENT_ERRORS)` around it) into its own file and a pure function and let every DB container which wants too implement their "wait for" check via SQLA just import it. Then we would have a simple single "DBContainer" without a dependency on SQLA, that one would have a abstract "_connect()` (that seems to be the name used by most of the other services; `_wait_for_ready()`would probably be more appropriate...).

So the DBContainer would be:

```py
from abc import ABC, abstractmethod
class DBContainer(ABC):
    @abstractmethod
    def _connect(self):
        """Check if the container is ready."""

    @abstractmethod
    def _configure(self):
        """Does basic setup so the container can start up (User, passwort, etc)."""

    @abstractmethod
    def get_connection_url(self, *, **kwargs): # psycopg v3 cannot handle sqla style URLs, so it needs a driver=None :-(
        ...
```

(although he first two could actually be part of the DockerContainer interface... Most containers have that method already and the rest does that stuff in `__init__`)

totallyzen:
> I haven't understood why this split is needed

I'm willing to hear "chill dude, it's fine". Please do call me out on my bias/pedanticism! I'm not averse to negative feedback. :slightly_smiling_face: 

The API level problem is what you are saying with `_wait_for_ready()` - this should be on all containers in fact, without fail.
I'm realising that maybe the "fix" to the API level problems with guarantees should live in it's own issue and merits it's own discussion, separate from this PR.

> pure function and let every DB container which wants too implement their "wait for" check via SQLA just import it. Then we would have a simple single "DBContainer" without a dependency on SQLA, that one would have to implement the abstract `_connect()` (that seems to be the name used by most of the other services; `_wait_for_ready()`would probably be more appropriate...) and one would be done.

In terms of dev experience, I think you are onto something! It would cause the least amount of hassle and focus on meaningful fixes rather than my personal pedantic issues with `generic.py`. It can also lean on eventually fixing the underlying missing guarantee of `_wait_for_ready()` being implemented for all containers (even if it's just `pass`)

Final note on breaking changes -> we have semver for this exact reason!
I strongly believe it's okay to have breaking changes as long as you communicate how to upgrade and tell the story to your users.
If we give them a clear "upgrade path" that has very few steps, we should be fine!
The frequency of breaking changes will also lower over time as we iron out these issues.

I tried to condense my thoughts, I hope this helps!

totallyzen:
To simplify your life for right now:

> Overall this is a BIG refactoring and it will touch probably half of the modules...

Aye, we can take it in two steps: 
- introduce the dirty fix in THIS PR -> I'll approve, just to clear out the issue of "too many dependencies"
- fix the `_wait_for_ready()` and `_configure()` abstract methods in a follow-up PR -> put it in 5.x

As for 5.x, we already have another important change that might break the API in the face of #314 which would put a time limit on container life (though I'm thinking how to implement this without breaking the API, if it's possible)

jankatins:
> >  I haven't understood why this split is needed
> I'm willing to hear "chill dude, it's fine". Please do call me out on my bias/pedanticism! I'm not averse to negative feedback. 🙂

Just putting this here in case of misunderstandings: that was a honest question, I've never seen such a layout and it took me a bit to understand how that fits together in the final package.

From my understanding (and after the merge of all code into one package) it seems that all code could live in a single `src/testcontainers/` folder if one is to accept that there is a `tests/` folder with the same (duplicated) hierarchy (so tests for `src/testcontainers/postgres` live in `tests/postgres`). Finding tests is a bit more scrolling (if your IDE does not let you jum to the tests) but packaging is probably a lot easier.

jankatins:
The dirty (and shirtes) fix is to 
- Remove the DepedencyFreeDBContainer class
- Overwrite _connect() in PG instead of `_verify_status()` WITHOUT calling super()

Because the current _connect() is importing in the funciton, not callign `super()._connect()` would be enough to not trigger the import.

This would also be a completely backwards compatible change (apart from using `testcontainers[postgres]` and not getting sqla installed, which is the actual feature :-))

jankatins:
Implmented the "quick and fast" version above.

Tested if locally by `poetry install --sync --extras=postgres` (which deinstalled sqla and psycopg)  and then running the tests `poetry run pytest -s --log-level=DEBUG modules/postgres`.

modules/postgres/tests/test_postgres.py
```diff
@@ -1,21 +1,19 @@
...
-def test_docker_run_postgres_with_driver_pg8000():
-    postgres_container = PostgresContainer("postgres:9.5", driver="pg8000")
-    with postgres_container as postgres:
-        engine = sqlalchemy.create_engine(postgres.get_connection_url())
-        with engine.begin() as connection:
-            connection.execute(sqlalchemy.text("select 1=1"))
+            status, msg = postgres.exec(f"psql -hlocalhost -p{postgres.port} -U{postgres.username} -c 'select 2*3*5*7*11*13*17 as a;' ")
+            assert "510510" in msg.decode("utf-8")
+            assert "(1 row)" in msg.decode("utf-8")
+            assert status == 0
```
jankatins, line 19: This still does not guarantee that the container is accessible form the outside, only from within the container :-( But doing that would need to install a driver in the dev dependencies or an additional tests dependency block :-(

alexanderankin:
this looks good to me. i understand there is a larger unfinished discussion about refactoring improvements - my thought is that if DbContainer was part of the public api (i would argue it is) then we cannot change it without version increases (maybe even major) so lets deprecate them now, and change the implementation underneath (that is something we are allowed to do on patch increases). (i just mean in the sense that people should not use DbContainer going forward if we plan to remove it.)

modules/postgres/tests/test_postgres.py
alexanderankin: we can install whatever dependencies in the dev group we want (unless im missing something)

alexanderankin: not sure why the old one needs to be removed:

```suggestion
            assert status == 0


def test_docker_run_postgres_with_sqlalchemy():
    postgres_container = PostgresContainer("postgres:9.5")
    with postgres_container as postgres:
        engine = sqlalchemy.create_engine(postgres.get_connection_url())
        with engine.begin() as connection:
            result = connection.execute(sqlalchemy.text("select version()"))
            for row in result:
                assert row[0].lower().startswith("postgresql 9.5")
```

jankatins: > we can install whatever dependencies in the dev group we want (unless im missing something)

One reason not to: to prove that the tests can run without them installed :-)

But yes, should be fine to do so.

pyproject.toml
```diff
@@ -102,7 +102,7 @@ neo4j = ["neo4j"]
 nginx = []
 opensearch = ["opensearch-py"]
 oracle = ["sqlalchemy", "cx_Oracle"]
-postgres = ["sqlalchemy", "psycopg2-binary"]
```
jankatins, line 105: Will move the psycopg2 dep to dev, too

jankatins: done

jankatins:
Added back the tests and also made a check against using sqla during container startup

jankatins:
> (i just mean in the sense that people should not use DbContainer going forward if we plan to remove it.)

After this discussion (with adding _wait_for_ready() and _configure() to the top level class), DBContainer would only contain this:

```
from abc import ABC, abstractmethod
class DBContainer(ABC):

    @abstractmethod
    def get_connection_url(self, *, **kwargs): # psycopg v3 cannot handle sqla style URLs, so it needs a driver=None :-(
        """Get a DB connection URL which can be used to connect to the DB."""

    def _create_connection_url( ... as now ...):
        """Helper to create a DB API connection URL."""
```

I'm not sure if this is worth a extra class n the hierarchy, especially because it's unclear to me if that URL can be useful for all DB container or only the ones which are based on the python DB API (=RDBMS ones, not stuff like mongodb or redis): I've no clue if thy also can be connected to with a single URL (kafka can not, I think). 

Given that each container would need to implement the method anyway, one could also make _create_connection_url() a pure function and let each container just declare the method themselves...

Anyway: I would appreciate if this conversation could be moved to a separate issue :-)

jankatins:
Added the fix to make ruff happy, waiting for workflow approval @alexanderankin @totallyzen

alexanderankin:
ill keep an eye on this today but also we have a slack https://slack.testcontainers.org/ where i respond to pings and stuff

jankatins:
And another try.. rand black and ruff locally, so hopefully no more rounds... @alexanderankin

alexanderankin:
ok lets iterate on this:)

PR #448, opened 2024-03-07
fix: Pin MongoDB images and improve test coverage for maintained versions

alexanderankin:
for some reason this causes an issue, see #401 for details


balint-backmaker:
Funny, never would have occurred to me to call the init function! I am not sure how pythonic it is to do this manually since `__init__` would have been called when you call `MongoClient()`

Could this be a version glitch where we didn't specify the version of the mongo dependency accurately enough, therefore it sucked up an unknown version?

(remind me to create an issue for dependency management too, we should install `renovate` on the repo sooner than later)

alexanderankin:
none of my research ever turned up anything to explain how this fixes the issue but @covatic-john claims that this fixes the issue in this comment - https://github.com/testcontainers/testcontainers-python/issues/401#issuecomment-1943956662

alexanderankin:
looking at this PR once a week and cannot bring myself to merge it. it makes no sense.

@covatic-john i did merge a different PR where it turns out there was no waiting at all for mongo container (i mean it usually starts in under 1sec in my experimence) so maybe it was that?

maybe i will change my mind one of these weeks. at present, we have no indication that the issue went away, so leaving open.

santi:
Having to call the `__init__` function twice points to some kind of race condition where the container was not ready before instantiating a client. I've revamped the MongoDB module with more tests and pinned versions in test images, for more reproducible behavior.

There are no traces of the behavior happening in version 5.0+ in any of the tests, so I'll merge this now and see if it persist. If it does, then we should look into the platform arch for the users with problems and see if it might be the same errors as in https://github.com/testcontainers/testcontainers-python/pull/457, where the connection string itself might be formatted incorrectly.

Issue #446, opened 2024-03-07
RFC: What to prioritise in 4.x releases?


totallyzen:
# Context

Now that we've got a release pipeline back in action, there is lots we can do to make things nicer.

I am raising this issue in order for community to gather and talk about what should be fixed as priorities and what improvements we want to add.

# Known Issues

We know there are several areas we want to improve:
- We want to thin out dependencies where possible - prefer giving a ready container but not be biased on how you should connect to the container. Example: postgres containers enforcing SqlAclhemy, as seen in #445 
- We want to improve package typing so people have better developer experiences, as seen in #305 and the PRs it spawned so far
- Adding new containers, as there is demand for it #439, #441, #389, etc

> [!NOTE]
> The list above is not conclusive, please comment on this issue and show your support!

max-pfeiffer:
@totallyzen  PS: I just checked for an M1/M2 GitHub runner. [It is actually available since January 2024.](https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/)

But it comes [without a Docker installation](https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md) and has other [quirks](https://github.com/actions/setup-python/issues/808). I just did a proof of concept with it: because of the mentioned issues I was not able to run the tests in the pipeline.

bearrito:
1. Revise documentation after the poetry workflow change. Most of the docs refer to the previous workflow.

2. Correct usage/mental model of `extras` and community modules. I get the desire for keeping dependencies low. I think the downside of not including a client dependency is while you can verify that the container is running, you can't verify that the connection details you might return are valid for the eventual client. Some want a full uri, others perhaps host and port. Or maybe I didn't expose the correct port etc.

 I'm the committer on https://github.com/testcontainers/testcontainers-python/pull/439 There is AFAIK on one supported python client to interact with NATS, so it seems natural to include it. I could take or leave any approach.

max-pfeiffer:
> Might be more appropriate to get that in a 5.x.
> 
> AFAIK ARM runners aren't generally available yet - https://resources.github.com/devops/accelerate-your-cicd-with-arm-and-gpu-runners-in-github-actions/ . Not sure you'd want to correlate ARM with MAC OS.
> 
> In general not too difficult to get that working with github actions using a matrix - https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#using-a-matrix-strategy

[...]

PR #460, opened 2024-03-09
docs(contributing): add contribution and new-container guide 

totallyzen:
# change

Document how to contribute, with initial focus on making local development smooth.

# Tasks
- [x] Finish the `new-container` guide
- [x] Remove any old docs referring to 
- [x] Update `README.md` to point at the contribution guide
- [x] Update `README.md` to add badges (supported python versions, etc) and to give kudos to current and past maintainers and contributors

PR #462, opened 2024-03-09
fix(nats): Client-Free(ish) NATS container

bearrito:
@totallyzen 
Closed previous PR because of dirty history and bad-merge.

Based of this comment - https://github.com/testcontainers/testcontainers-python/issues/446#issuecomment-1986860683 
I removed hard dependency on the NATS client.

This PR does not require a client to work or validate at a basic level.
However if a developer wants to use a client, there is test support for that.


modules/nats/tests/test_nats.py
```diff
@@ -0,0 +1,101 @@
...
+If you are developing this and you want to test more advanced scenarios using a client
+Activate your poetry shell.
+pip install nats-py
+This will get nats-py into your environment but keep it out of the project
+
+
+"""
+
+
+NO_NATS_CLIENT = True
```
bearrito, line 16: This section will check if you have the client available, if so then certain tests are available. If not then those tests are skipped.

modules/nats/testcontainers/nats/__init__.py
```diff
@@ -0,0 +1,65 @@
+#
```
bearrito, line 1: No client imported in this file.

modules/nats/tests/test_nats.py
```diff
@@ -0,0 +1,101 @@
...
+    Returns:
+        client: Nats client to connect to the container.
+    """
+    conn_string = container.nats_uri()
+    client = await nats_connect(conn_string)
+    return client
+
+
+def test_basic_container_ops():
+    with NatsContainer() as container:
```
bearrito, line 39: This tests runs no matter what.

modules/nats/tests/test_nats.py
```diff
@@ -0,0 +1,101 @@
...
+        uri = container.nats_uri()
+        management_uri = container.nats_management_uri()
+
+        assert uri != management_uri
+
+
+pytest.mark.usefixtures("anyio_backend")
+
+
+@pytest.mark.skipif(NO_NATS_CLIENT, reason="No NATS Client Available")
```
bearrito, line 53: Skipped if no client.

alexanderankin:
@bearrito for what its worth we havent decided what we are doing with the dependencies yet - i was going to merge your PR as is after releasing whatever other fixes we can fit before we add features (things that have to bump minor version)

modules/nats/tests/test_nats.py
alexanderankin: there's really no reason to exclude tests due to missing deps - just dont force a dep on the client code which is using the testcontainer

bearrito:
@alexanderankin No worries. If you guys decide to allow more dependencies. I can add to the extras and then remove the try/catch import

modules/nats/tests/test_nats.py
bearrito: What's the correct way to indicate that with poetry?

totallyzen:
@bearrito my take at this point is that you can add it as a dev dependency - the point is more so to prevent enforcement of clients, not to make testing a hellscape.

I've reflected this in the latest changes on #460 as well, comment there instead on this if you agree!

alexanderankin:
what happened here? i am rebasing all the branches because too crazy otherwise

bearrito:
@alexanderankin  I think I was in the middle of you doing that. What do you want me to do here?

alexanderankin:
@bearrito no problem, i just merged all the newer ones, about to rebase this one and merge it. thanks for your contribution and support here. let me know if you want to have a final look over in a few mins before i push and hit merge.

alexanderankin:
need to fix the doctest still, one min

modules/nats/tests/test_nats.py
alexanderankin: > What's the correct way to indicate that with poetry?

extremely painfull :smile:

PR #463, opened 2024-03-11
fix: Qdrant module

Anush008:
This PR adds a module to spawn a [Qdrant](https://qdrant.tech) test container.


kiview:
Thanks for the PR @Anush008 🙌 
There seems to be an issue with the docstrings:
https://github.com/testcontainers/testcontainers-python/actions/runs/8247542755/job/22555998514?pr=463#step:5:76

modules/qdrant/testcontainers/qdrant/__init__.py
```diff
@@ -0,0 +1,145 @@
...
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+import os
+from functools import cached_property
+from pathlib import Path
+from typing import Optional
+
+from qdrant_client import AsyncQdrantClient, QdrantClient
```
eddumelendez, line 18: Can we please avoid coupling to the qdrant API?

Anush008: I added this based on the other container modules.
https://github.com/testcontainers/testcontainers-python/blob/5758310532b8a8e1303a24bc534fa8aeb0f75eb2/modules/mongodb/testcontainers/mongodb/__init__.py#L87

https://github.com/testcontainers/testcontainers-python/blob/5758310532b8a8e1303a24bc534fa8aeb0f75eb2/modules/neo4j/testcontainers/neo4j/__init__.py#L73

alexanderankin: I think we will circle back later to remove all of these - its too much to think about doing all up front. in the future i think we'll actually move the imports into the offending methods and surround with try except ImportError and re-raise with a message telling the user to install the module and not use the methods.

Anush008: Turned them to optional imports.

https://github.com/testcontainers/testcontainers-python/pull/463/commits/e9354f19577242ad6ee6907da2274edb3790ef43

PR #167, opened 2021-12-30
feat: Adding support for Cassandra and Scylla

fruch:
This add the support for those cassandra based dbs
and their drivers, cassandra-driver, scylla-driver

Ref: https://cassandra.apache.org/
Ref: https://www.scylladb.com/
Ref: https://pypi.org/project/cassandra-driver/
Ref: https://pypi.org/project/scylla-driver/

PR #476, opened 2024-03-14
fix: Add CassandraContainer

cbornet:


modules/cassandra/testcontainers/cassandra/__init__.py
```diff
@@ -0,0 +1,53 @@
...
+            ...    result = session.execute("SELECT release_version FROM system.local;")
+    """
+
+    def __init__(self, image: str = "cassandra:latest", port: int = 9042, **kwargs) -> None:
+        super().__init__(image=image, **kwargs)
+        self.port = port
+        self.with_exposed_ports(self.port)
+
+    def _connect(self):
+        wait_for_logs(self, "Startup complete")
```
cbornet, line 48: What is the preferred strategy for the readiness probe ?
I have used `wait_for_logs` here because it doesn't need to depend on the cassandra driver.
But I see that the Java implementation [uses the driver](https://github.com/testcontainers/testcontainers-java/blob/main/modules/cassandra/src/main/java/org/testcontainers/containers/wait/CassandraQueryWaitStrategy.java#L33) and it gives the opportunity to [provide an init script](https://github.com/testcontainers/testcontainers-java/blob/main/modules/cassandra/src/main/java/org/testcontainers/containers/CassandraContainer.java#L154).

alexanderankin:
there is working on going here https://github.com/testcontainers/testcontainers-python/pull/167#issuecomment-1996362694 too

cbornet:
Oh, sorry I missed that ! The PR has been opened for more than 2 years ?!
Feel free to merge any of them. With your latest https://github.com/fruch/testcontainers-python/commit/2b238a227b00445bdef42427f1504fc87081800e they are very similar.
Some differences to note:
* This one doesn't depend on the cassandra driver. It exposes a contact points tuple instead of a Cluster(). The tuple can be fed directly to a Cluster constructor.
* This one adds an entry in .github/settings.xml

PR #478, opened 2024-03-15
fix(core): Stop container when signals captured

lamcw:
The current `atexit` hook that stops (and optionally remove) containers in the Docker daemon would not work because of limitations of the `atexit` module: the handlers are not executed if the Python interpreter is

> killed by a signal not handled by Python, when a Python fatal internal error is detected, or when [os._exit()](https://docs.python.org/3/library/os.html#os._exit) is called.
> https://docs.python.org/3/library/atexit.html

To handle shutdowns of containers completely (for e.g. when the tests are terminated via SIGINT / SIGTERM), we'd need to register signal handlers for them as well.

This kind of acts as an interim solution while #314 is being implemented.

### DIFF
--- a/core/testcontainers/core/docker_client.py
+++ b/core/testcontainers/core/docker_client.py
@@ -13,10 +13,11 @@
 import atexit
 import functools as ft
 import os
+import signal
 import urllib
 from os.path import exists
 from pathlib import Path
-from typing import Optional, Union
+from typing import Optional, Union, Callable, Iterable, Optional
 
 import docker
 from docker.errors import NotFound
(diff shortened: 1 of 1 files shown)


core/testcontainers/core/docker_client.py
```diff
@@ -154,3 +160,28 @@ def _stop_container(container: Container) -> None:
         pass
     except Exception as ex:
         LOGGER.warning("failed to shut down container %s with image %s: %s", container.id, container.image, ex)
+
+
+def _register_signal_handler(handler: Callable[[], None], signals: Iterable) -> None:
```
lamcw, line 165: A more advanced version of `signal.signal` that does not overwrite existing signal handler registered to signal.

santi:
This issue will be resolved when https://github.com/testcontainers/testcontainers-python/pull/314 lands, which removes the `atexit` hook completely and removes the dependency on GC to cleanup danglings containers

alexanderankin:
going to close since reaper v1 has landed, and even reaper v2 (only difference being `auto_remove=True` (aka `docker run --rm`) so the instances dont hang out and haunt your system :ghost:)

if any other issues with stopping containers or anything else, just give a shout! @lamcw

PR #479, opened 2024-03-15
fix(arangodb): tests to pass on ARM CPUs - change default image to 3.11.x where ARM image is published

max-pfeiffer:
- updated arangodb image version which now supports ARM CPUs
- skipping test for legacy version on ARM CPU

fixes https://github.com/testcontainers/testcontainers-python/issues/449

Screenshot 2024-03-15 at 11 43 35

### DIFF
--- a/modules/arangodb/tests/test_arangodb.py
+++ b/modules/arangodb/tests/test_arangodb.py
@@ -7,8 +7,10 @@ from arango import ArangoClient
 from arango.exceptions import DatabaseCreateError, ServerVersionError
 
 from testcontainers.arangodb import ArangoDbContainer
+import platform
 
 ARANGODB_IMAGE_NAME = "arangodb"
+IMAGE_VERSION = "3.11.8"
 
 
 def arango_test_ops(arango_client, expeced_version, username="root", password=""):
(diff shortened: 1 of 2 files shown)


modules/arangodb/tests/test_arangodb.py
```diff
@@ -100,8 +101,7 @@ def test_docker_run_arango_random_root_password():
     """
     Test ArangoDB container with ARANGO_RANDOM_ROOT_PASSWORD var set.
     """
-    image_version = "3.9.1"
-    image = f"{ARANGODB_IMAGE_NAME}:{image_version}"
+    image = f"{ARANGODB_IMAGE_NAME}:{IMAGE_VERSION}"
```
santi, line 104: Since this is now only using version 3.11.x, maybe we can remove this test and instead only have a single test for the 3.11 version that works on both x86 and ARM architecture?

Or do we have a strategy on supporting official modules with older versions? @balint-backmaker @alexanderankin @kiview

santi: We have 4 tests in this file where it checks different combinations of passwords and an older version of the container image. Part of the testing also includes inserting data and checking that you get the desired data out again.

I don't think testing queries outside of a simple version check on the `_system` table should be in scope for Testcontainers, that is better suited for testing in the client package itself. Supplying the wrong password as part of connection string is also not really the scope of this project

Therefore I suggest:
- Remove `test_docker_run_arango_without_auth()` and `test_docker_run_arango_random_root_password()`, then combine `test_docker_run_arango()` with `arango_test_ops()` and remove logic related to testing queries and DB setup other than checking the `_system` database for the current version.
- Keep the old version test for now. Doesn't really hurt having it here for documented backwards compatibility

alexanderankin: I really dont see a reason to change anything more than necessary here, now that we have filters to only run module specific tests when something in the module changes.

(sorry for radio silence, was stressed last week, then got sick (probably not a coincidence; still sick).

PR #480, opened 2024-03-15
fix(keycloak): tests on aarch64, use image from [jboss -> quay], change supported version [16+ -> 18+]

max-pfeiffer:
- jboss/keycloak is discontinued, adapting the official Docker image: https://quay.io/repository/keycloak/keycloak
- switched to the latest image version with ARM support

fixes https://github.com/testcontainers/testcontainers-python/issues/451
fixes https://github.com/testcontainers/testcontainers-python/issues/483

Screenshot 2024-03-15 at 12 42 29

### DIFF
--- a/modules/keycloak/tests/test_keycloak.py
+++ b/modules/keycloak/tests/test_keycloak.py
@@ -1,9 +1,6 @@
-import pytest
-
 from testcontainers.keycloak import KeycloakContainer
 
 
-@pytest.mark.parametrize("version", ["16.1.1"])
-def test_docker_run_keycloak(version: str):
-    with KeycloakContainer(f"jboss/keycloak:{version}") as kc:
-        kc.get_client().users_count()
+def test_docker_run_keycloak():
+    with KeycloakContainer("quay.io/keycloak/keycloak:24.0.1") as keycloak_admin:
+        keycloak_admin.get_client().users_count()
--- a/modules/keycloak/testcontainers/keycloak/__init__.py
+++ b/modules/keycloak/testcontainers/keycloak/__init__.py
@@ -36,20 +36,21 @@ class KeycloakContainer(DockerContainer):
 
     def __init__(
         self,
-        image="jboss/keycloak:latest",
+        image="quay.io/keycloak/keycloak:latest",
         username: Optional[str] = None,
         password: Optional[str] = None,
         port: int = 8080,
     ) -> None:
         super().__init__(image=image)
-        self.username = username or os.environ.get("KEYCLOAK_USER", "test")
-        self.password = password or os.environ.get("KEYCLOAK_PASSWORD", "test")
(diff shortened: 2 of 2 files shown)


modules/keycloak/tests/test_keycloak.py
```diff
@@ -3,7 +3,7 @@
 from testcontainers.keycloak import KeycloakContainer
 
 
-@pytest.mark.parametrize("version", ["16.1.1"])
+@pytest.mark.parametrize("version", ["24.0.1"])
```
santi, line 6: Since we only test for a single parameter, you can probably just inline this version

max-pfeiffer: @santi  Thanks for your feedback. I thought about this already myself. Now we already two. 😃 I'll scratch that parametrization.

matheusvnm:
@max-pfeiffer, allow me to give my 2 cents: The new Keycloak version has a specific endpoint for readiness checks (`/health/ready`). It can be activated by setting the 'KC_HEALTH_ENABLED' to`true`. Why not use it to reliably check the state of the system?

You could do something like below:

```python3
def _configure(self):
    self.with_env('KEYCLOAK_ADMIN', self.KEYCLOAK_USER)
    self.with_env('KEYCLOAK_ADMIN_PASSWORD', self.KEYCLOAK_PASS)
    self.with_env('KC_HEALTH_ENABLED', 'true')
    self.with_command('start-dev')

@wait_container_is_ready(requests.exceptions.ConnectionError, requests.exceptions.ReadTimeout)
def _connect(self):
    response = requests.get(f"{self.get_url()}/health/ready", timeout=1)
    response.raise_for_status()
```

max-pfeiffer:
@matheusvnm Good point!  I was not aware that these endpoints became introduced. I agree: this would be a better way to implement a readiness probe.

Problem is here: it looks like these endpoints became introduced with v18.0. So we would cut off all the older version which people might still be using.

But we also want to have ARM support here. Images with ARM64 support are available since v17.0, I just checked the [container registry](https://quay.io/repository/keycloak/keycloak?tab=tags&tag=latest).

So supporting versions from 18.0 to 24.0.1 could be an acceptable trade off.

santi:
Keycloak themselves only support the [latest major version](https://www.keycloak.org/security.html#:~:text=googlegroups.com.-,Supported%20Versions,-Depending%20on%20the), so shrinking the supported version range to 18.x+ sounds fine by me.

alexanderankin:
if anyone really needs ancient versions im sure we can put together some kind of code sample with extending the core class `DockerContainer` with whatever they need, but it doesn't need to be in the library that gets published to pypi. I think we have enough of a consensus on that to merge? i even remember seeing something about trying to get keycloak to maintain their own testcontainers, so im not trying to overthink this.

Issue #489, opened 2024-03-21
Bug: ryuk exited instances leaking


shoffmeister:
**Describe the bug**

With the new ryuk container usage in v4.1.1, dependent containers get cleaned up nicely, but the ryuk container itself remains, in exited state. The ryuk container itself should also be cleaned up.

`docker ps --all` will show a great many of these containers, see the very end of the bug report for some testcontainers-based pytest integration test that was running in a loop.

While `docker rm -f $(docker ps --all --quiet --filter="name=testcontainers-ryuk*")` will clean this up, it is still annoying to have to manage this.

**To Reproduce**

Any usage of testcontainers should exhibit this behaviour.

Locally, I am running this with a specialized Dockercontainer implementation which drives the official Apache Kafka 3.7.0 container image in KRaft mode (which obviates the need for Zookeeper dancing and hence cycles much much faster).

**Runtime environment**

Provide a summary of your runtime environment. Which operating system, python version, and docker version are you using? What is the version of `testcontainers-python` you are using? You can run the following commands to get the relevant information.

```bash
# Get the operating system information (on a unix os).
$ uname -a

Linux arch 6.8.1-arch1-1 #1 SMP PREEMPT_DYNAMIC Sat, 16 Mar 2024 17:15:35 +0000 x86_64 GNU/Linux

# Get the python version.
$ python --version

Python 3.11.8

# Get the docker version and other docker information.
$ docker info

❯ docker info
Client:
 Version:    25.0.4
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  0.13.1
    Path:     /usr/lib/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  2.24.7
    Path:     /usr/lib/docker/cli-plugins/docker-compose

Server:
 Containers: 50
  Running: 0
  Paused: 0
  Stopped: 50
 Images: 34
 Server Version: 25.0.4
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: true
  Native Overlay Diff: false
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
[...]

PR #491, opened 2024-03-23
fix: clean up reaper instances

alexanderankin:
suggest a fix for #489

### DIFF
--- a/core/tests/test_ryuk.py
+++ b/core/tests/test_ryuk.py
@@ -22,3 +22,8 @@ def test_container_without_ryuk(monkeypatch):
     with DockerContainer("hello-world") as cont:
         wait_for_logs(cont, "Hello from Docker!")
         assert Reaper._instance is None
+
+
+def test_reaper_gets_deleted():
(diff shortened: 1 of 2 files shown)


alexanderankin:
can you test with https://pypi.org/project/testcontainers/4.2.0rc1/ ? thanks!

alexanderankin:
superceded by #499

core/tests/test_ryuk.py
```diff
@@ -1,9 +1,14 @@
...
+
+from contextlib import contextmanager
+
 from testcontainers.core import container
 from testcontainers.core.container import Reaper
 from testcontainers.core.container import DockerContainer
 from testcontainers.core.waiting_utils import wait_for_logs
 
 
+@pytest.mark.skip("invalid test - ryuk logs 'Removed' right before exiting")
```
kiview, line 11: Why don't we remove the test if it is invalid?
I think the test should check, that the Ryuk container and the Docker container are gone in this case?

Issue #482, opened 2024-03-15
Bug: tests are run inconsistently in CI pipeline


max-pfeiffer:
**Describe the bug**

Maybe not a real bug but more an issue I came across when I looked at how our tests are run in CI pipeline:

- core: here we have a [matrix covering Python versions from 3.9 - 3.12](https://github.com/testcontainers/testcontainers-python/tree/main/.github/workflows)
- community modules: we have a [matrix without real purpose covering only Python version 3.11](https://github.com/testcontainers/testcontainers-python/blob/main/.github/workflows/ci-community.yml)

So looking at pyproject.toml I guess we want to support Python versions from 3.9 - 4.0, so our pipeline needs to reflect this.

I also see that we do not pin the versions of our Ubuntu runners. I would say it's a better practice to pin version to get predictable CI results.

Also I am scratching my head what purpose that make file still serves there. I see no need to use it in the pipeline jobs.

Also we use coverage, but are not doing anything with that test coverage data. 

I could spend some time fixing this up. But I would be very grateful for some feedback and thoughts. Not so sure if I miss something here.

max-pfeiffer:
@totallyzen Ok, so I will create a PR addressing these inconsistency issues and pin runner versions.

I just went through https://github.com/testcontainers/testcontainers-python/pull/460 and left some comments and suggestions.

I haven't used slack in ages. But I will check on that channel now.

PR #497, opened 2024-03-24
fix: inconsistent test runs for community modules

max-pfeiffer:
- Fixed inconsistencies for community module test runs: using all supported Python versions
- Pinned runner version (best practice)

fixes https://github.com/testcontainers/testcontainers-python/issues/482

### DIFF
--- a/.github/workflows/ci-community.yml
+++ b/.github/workflows/ci-community.yml
@@ -14,7 +14,7 @@ on:
 
 jobs:
   track-modules:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
       - name: Checkout contents
         uses: actions/checkout@v4
@@ -38,14 +38,14 @@ jobs:
     outputs:
       changed_modules: ${{ steps.compute-changes.outputs.computed_modules }}
   test:
+    runs-on: ubuntu-22.04
     needs: [track-modules]
     if: ${{ needs.track-modules.outputs.changed_modules != '[]' }}
     strategy:
       fail-fast: false
       matrix:
-        python-version: [ "3.11" ]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
         module: ${{ fromJSON(needs.track-modules.outputs.changed_modules) }}
-    runs-on: ubuntu-latest
     steps:
       - name: Checkout contents
         uses: actions/checkout@v4
--- a/.github/workflows/ci-core.yml
+++ b/.github/workflows/ci-core.yml
@@ -10,12 +10,11 @@ on:
 
 jobs:
   test:
+    runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu ]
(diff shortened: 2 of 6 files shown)


.github/workflows/ci-community.yml
```diff
@@ -38,14 +38,14 @@ jobs:
     outputs:
       changed_modules: ${{ steps.compute-changes.outputs.computed_modules }}
   test:
+    runs-on: ubuntu-22.04
     needs: [track-modules]
     if: ${{ needs.track-modules.outputs.changed_modules != '[]' }}
     strategy:
       fail-fast: false
       matrix:
-        python-version: [ "3.11" ]
```
alexanderankin, line 46: im prett sure we did this as a cost saving measure

.github/workflows/ci-core.yml
```diff
@@ -10,12 +10,11 @@ on:
...
+    runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu ]
         python-version: ["3.9", "3.10", "3.11", "3.12"]
-    runs-on: ${{ matrix.os }}-latest
     steps:
       - uses: actions/checkout@v4
       - name: Set up Python
```
alexanderankin, line 20: i think this is also aspirationally there for win/mac OSs, but again they are missing as a cost saving measure.

alexanderankin (commented):
this is not arguing for or against, im just documenting the previous rationale.

i think making it the ubuntu version explicit may be justifiable. we have a follow up in the short term, as ubuntu 24 is just around the corner.

.github/workflows/ci-community.yml
max-pfeiffer: @alexanderankin Please check what the job track-modules does: it tracks code changes in the modules and then runs the test job conditionally. So this is only executed when we actually have relevant code changes. 

So I would say that job is already quite optimised with regard to pipeline usage time.

Testing with the Python versions which we actually want to support is kind of mandatory if you want to avoid our package breaking.

What would you suggest doing to cover these test cases?

.github/workflows/ci-core.yml
max-pfeiffer: @alexanderankin Please be aware of the findings from @bearrito and me concerning ARM runners in this thread: https://github.com/testcontainers/testcontainers-python/issues/446

It looks like these runners are usable mid or long term. Maybe end of this year. We don't know.

Currently this matrix does not serve any purpose. So I consider this as dead code which can be removed.

alexanderankin:
ok I'm going to merge because i guess it seems reasonable enough.

pinging for FYI/attention in case any issues come up with this change:

* @kiview
* @totallyzen  
* @santi

.github/workflows/ci-community.yml
santi: A bit late to the party here, but fully support testing on all Python versions to be sure about compatibility. If we ever run into GA minutes becoming scarce, we can have a conversation about cost savings then

PR #504, opened 2024-03-25
fix: typing in core

alexanderankin:


totallyzen:
Nice! Glad you found the time before me! Good start! :pray:

core/testcontainers/core/docker_client.py
```diff
@@ -98,36 +98,36 @@ def find_host_network(self) -> Optional[str]:
...
         Lookup the public-facing port that is NAT-ed to :code:`port`.
         """
         port_mappings = self.client.api.port(container_id, port)
         if not port_mappings:
             raise ConnectionError(f"Port mapping for container {container_id} and port {port} is " "not available")
-        return port_mappings[0]["HostPort"]
+        return cast(str, port_mappings[0]["HostPort"])
 
-    def get_container(self, container_id: str) -> Container:
+    def get_container(self, container_id: str) -> dict[str, Any]:
```
jankatins, line 115: This looks like a regression to me.

(I would argue that nothing should return a dict directly, ever, but always a typed dictionary, dataclass or something similar with described fields. In my experience, interfaces with dict always end in maintenance headache)

core/testcontainers/core/waiting_utils.py
```diff
@@ -78,7 +78,10 @@ def wait_for(condition: Callable[..., bool]) -> bool:
 
 
 def wait_for_logs(
-    container: "DockerContainer", predicate: Union[Callable, str], timeout: Optional[float] = None, interval: float = 1
+    container: "DockerContainer",
+    predicate: Union[Callable[..., Any], str],
```
jankatins, line 82: Shouldn't the callable here be `Callable[(str,) bool]`? It's used as `if predicate(stdout) or predicate(stderr):`

core/testcontainers/core/docker_client.py
alexanderankin: in the opinion of the project, it is also a maintenance burden to return types from other libraries. either way, I think breaking the actual return and changing runtime behavior would be worse than fixing the type.

core/testcontainers/core/waiting_utils.py
alexanderankin: in java it also takes regex i dont remember if that is the case here but i will leave this thread open to review when and if this PR goes on. I think it belongs in a draft state for now. i just wanted to record the results of going through the exercise to add types. for example, some of the places we return ports are actually `str | int` (well, union, for `3.8` compatibility)

alexanderankin:
closed in favor of #810

PR #506, opened 2024-03-26
chore(build): improve devcontainers setup with poetry

bearrito:
1. Replaces uses an image directly.
2. Creates a custom image, still based off the original Image.
3. Installs poetry inside the container. This saves the user about 20 seconds every container load.
4. Installs `pre-commit` in the container. Saves more time.
5. Enables `pre-commit`. From inside the container the user cannot commit without `pre-commit` running.
6. Adds in git autocomplete to the container (see screenshot)
7. Adds in poetry autocomplete to the container (see screenshot)
devcontainer-completion


alexanderankin:
i tested my new approach but i think it makes sense for you to see if it works for you, given that i dont use this all that much and may not be testing it the same way.

bearrito:
@alexanderankin No this is good, Much cleaner. I really just needed a way to get the completions in there.

totallyzen (approved):
LGTM, thank you! :heart:

PR #507, opened 2024-03-26
feat(client): Add custom User-Agent in Docker client as `tc-python/<version>`

eddumelendez:
Set User-Agent in format `tc-python/<version>`, `version` value is coming from `pyproject.toml`.

The `User-Agent` header will allow to identify Testcontainers language implementation and the specific version. Also, track the usage of the library.


santi (approved):
Nicely done 🙌

PR #508, opened 2024-03-26
fix(google): add support for Datastore emulator

MattOates:
Expands the google module with a DatastoreContainer using the beta Datastore emulator using the same image as the PubSubContainer.

Im already using a local copy of this in production. It would be nice to not have to support copy paste solutions and instead see it added to the google module. This is my first PR so please let me know what I need to do to get this over the line. Thanks.

Looks like @tillahoffmann wrote the original PubSub emulator container


alexanderankin:
https://testcontainers.readthedocs.io/en/latest/modules/google/README.html

PR #363, opened 2023-06-11
fix: Improved Oracle DB module

gvenzl:
Hi,

I took the liberty to improve the Oracle DB module for Testcontainers Python. The PR has several enhancements:

* Leveraging `oracledb` thin Python driver
    * This makes Oracle DB tests on CI/CD now possible too
* Usage of `gvenzl/oracle-free` image with the latest and greatest Oracle DB version
* DB version independent readiness check
* Support for various `gvenzl/oracle-free` image features (`ORACLE_DATABASE`, `APP_USER`, `APP_USER_PASSWORD`, etc)
* Tests for Oracle DB for the various combinations

Ideally, some more documentation on how the Container is supposed to be used would be handy but I couldn't really find a good example of how such a ReadMe should be structured.
Any things are gladly appreciated!

PR #513, opened 2024-03-30
fix(oracle): Remove unmaintained Oracle community module

santi:
There are multiple problems with the current support for the Oracle module. 
- The author of the image seems to [no longer maintain it](https://github.com/wnameless/docker-oracle-xe-11g)
- The original image (without the `-r2` suffix was DMCA'd by Oracle from DockerHub. Not sure if the new one is actually compliant or if it has just gone under the radar.
- The image uses an old build of Ubuntu 18.04 (which reached EOL in May 2023)
- There is really nothing special in the setup and running of the image, which would not qualify for a community module currently (no wait strategy, almost no custom setup). An example code snippet would suffice in this case.
- It does not support running on Apple Silicone, and requires client libraries to be installed to run in general.

Therefore, I suggest that we remove this module.


santi:
Just discovered this one: https://github.com/testcontainers/testcontainers-python/pull/363 Will check out it and similar ideas before moving forward with this one.

totallyzen (approved):
As a general concept, if #363 doesn't work out, you've got my approval to remove. It's better to keep a healthy project than broken features.

This also comes back to "what level of guarantee do we have for community modules?" with this being a breaking change (or not) - either way, I approve the concept, but may want to wait with merging it for the correct semver identification

:pray:

alexanderankin (changes requested):
we have a PR to update oracle to mr. venzl's version, if we merge this, it gets harder to rebase that PR, please consider not merging, as i do have a plan for this module.

alexanderankin:
sorry, just saw the other newly opened PR, catching up still...

alexanderankin:
not necessary after #363

PR #514, opened 2024-03-30
feat(oracle): Use Oracle Free image in Oracle module

santi:
There are multiple problems with the current support for the Oracle module.

- The author of the image seems to [no longer maintain it](https://github.com/wnameless/docker-oracle-xe-11g)
- The original image (without the -r2 suffix was DMCA'd by Oracle from DockerHub. Not sure if the new one is actually compliant or if it has just gone under the radar.
- The image uses an old build of Ubuntu 18.04 (which reached EOL in May 2023)
- There is really nothing special in the setup and running of the image, which would not qualify for a community module currently (no wait strategy, almost no custom setup). An example code snippet would suffice in this case.
- It does not support running on Apple Silicone, and requires client libraries to be installed to run in general.

By switching over to Oracle Free image as our base image, we get an up-to-date image which is maintained on a modern Oracle version.


modules/mysql/tests/test_mysql.py
```diff
@@ -30,6 +30,17 @@ def test_docker_run_mysql_8():
...
+def test_docker_run_mysql_8():
+    config = MySqlContainer("mysql:8")
+    with config as mysql:
+        engine = sqlalchemy.create_engine(mysql.get_connection_url())
+        with engine.begin() as connection:
+            result = connection.execute(sqlalchemy.text("select version()"))
+            for row in result:
+                assert row[0].startswith("8")
+
+
```
totallyzen, line 43: probably not intended, it touches `mysql` :slightly_smiling_face:

pyproject.toml
```diff
@@ -78,7 +78,7 @@ pymssql = { version = "*", optional = true }
 pymysql = { version = "*", extras = ["rsa"], optional = true }
 neo4j = { version = "*", optional = true }
 opensearch-py = { version = "*", optional = true }
-cx_Oracle = { version = "*", optional = true }
+oracledb = { version = "*", optional = true }
```
santi, line 81: This is only needed during testing, so should be included as a dev dependency, not as part of the `extra`.

pyproject.toml
```diff
@@ -101,7 +101,7 @@ mysql = ["sqlalchemy", "pymysql"]
 neo4j = ["neo4j"]
 nginx = []
 opensearch = ["opensearch-py"]
-oracle = ["sqlalchemy", "cx_Oracle"]
+oracle = ["sqlalchemy", "oracledb"]
```
santi, line 104: Same as [here](https://github.com/testcontainers/testcontainers-python/pull/514/files#r1545495696), should only be a dev dependency (so that we don't include a lot of libs when installing TC)

modules/mysql/tests/test_mysql.py
santi: Yupp. I don't have push to David's repo, but I took the liberty of opening a PR on a fork that David got from https://github.com/testcontainers/testcontainers-python/pull/363.

santi:
@alexanderankin I took the liberty of opening a PR from your fork, as I didn't want to copy your work when you were almost at the finish line. Just a few comments regarding MySQL and some dependencies, and you have a thumbs up from me 👍 Judging from some reactions over at https://github.com/testcontainers/testcontainers-python/pull/363, this is a useful module

modules/mysql/tests/test_mysql.py
alexanderankin: it is also unintended because it is an exact copy of the preceding lines

alexanderankin:
closing in favor of #363

PR #518, opened 2024-04-01
fix(cassandra): make cassandra dependency optional/test-only

alexanderankin:
image


alexanderankin:
actually after a careful review of the docs, im fairly sure this does what is desired.

PR #520, opened 2024-04-02
feat: Run GA pipelines on MacOS 14 (M1 arch)

santi:


santi:
Closed due to hardware virtualization limits on MacOS M1 runners by Github. https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/about-larger-runners#limitations-for-macos-larger-runners

Until a runner with M2+ architecture is released or another ARM64-runner is available, testing on ARM64 is not possible without heavy virtualization, which would make the test suite take forever to complete.

PR #523, opened 2024-04-02
fix(core): Improving typing for common container usage scenarios

santi:
Improves type hints for type checking in a common use cases:

```python
with MySqlContainer("mysql:8").with_env("some", "value") as mysql:
    url = mysql.get_connection_url()  # get_connection_url would previously be an unknown member here
```
And, also improved type hinting for the custom `DockerClient`'s `run` command, where the linter no longer reports an error due to missing parameter types:
```python
DockerClient.run("nginx")  # Previously this would report "Argument missing for parameter "image"
```

### DIFF
--- a/core/testcontainers/core/container.py
+++ b/core/testcontainers/core/container.py
@@ -54,29 +54,29 @@ class DockerContainer:
         self._name = None
         self._kwargs = kwargs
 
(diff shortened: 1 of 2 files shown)


alexanderankin (approved):
'spose

alexanderankin:
linking #504 and #305

PR #525, opened 2024-04-03
fix(dependencies): remove usage of `sqlalchemy` in DB extras. Add default wait timeout for `wait_for_logs`

santi:
Removes usage of `sqlalchemy`, as part of the work described in https://github.com/testcontainers/testcontainers-python/issues/526.

- Adds default timeout to the `wait_for_logs` waiting strategy, the same timeout used by default in the `wait_container_is_ready` strategy.
- Changes wait strategy for `mysql` container to wait for logs indicating that the DB engine is ready to accept connections (MySQL performs a restart as part of its startup procedure, so the logs will always appear twice.
- Add More tests for different `mysql` and `mariadb` versions to ensure consistency in wait strategy.
- Remove x86 emulation for ARM devices for MariaDB, as it MariaDB images support ARM architectures already.
- Change wait strategy for `oracle-free`, as the images produce a consistent `DATABASE IS READY TO USE!` log message on startup.

Next steps will be to remove `sqlalchemy` as a bundled dependency entirely, but I have not included it in this PR as I consider it a bigger change than just changing wait strategies as an internal implementation detail. I plan to do this as part of a bigger rework where i remove the `DbContainer` class and standardize configuration hooks and wait strategies across containers (not just DB containers, all containers in need of a configuration and readiness step). See https://github.com/testcontainers/testcontainers-python/pull/527 for WIP.


alexanderankin:
since this is such a small change to core and the pypi semver refers to core not modules, ive changed this to a fix. i think that makes sense and what was discussed here - https://github.com/testcontainers/testcontainers-python/issues/446#issuecomment-2027901238 - if that is okay, please merge @santi

alexanderankin (approved):

Issue #459, opened 2024-03-09
what to do with DBContainer - split from removing sqlalchemy dependency


alexanderankin:
from https://github.com/testcontainers/testcontainers-python/pull/445#issuecomment-1985436334

> After this discussion (with adding _wait_for_ready() and _configure() to the top level class), DBContainer would only contain this:
> ...
> Anyway: I would appreciate if this conversation could be moved to a separate issue :-)

jankatins:
From #445 

```py
from abc import ABC, abstractmethod
class DBContainer(ABC):

    @abstractmethod
    def get_connection_url(self, *, **kwargs): # psycopg v3 cannot handle sqla style URLs, so it needs a driver=None :-(
        """Get a DB connection URL which can be used to connect to the DB."""

    def _create_connection_url( ... as now ...):
        """Helper to create a DB API connection URL."""
```

I'm not sure if this is worth a extra class in the hierarchy, especially because it's unclear to me if that URL can be useful for all DB container or only the ones which are based on the python DB API (=RDBMS ones, not stuff like mongodb or redis): I've no clue if they also can be connected to with a single URL (kafka can not, I think). 

Given that each container would need to implement the method anyway, one could also make `_create_connection_url()` a pure function and let each container just declare the method themselves...

PR #527, opened 2024-04-04
refactor(core): Standardize configuration and readiness steps in container lifecycle

santi:
Aims for this PR:
- Include `_configure()` and `_wait_until_ready()` hooks in base `DockerContainer` image to be used by all container implementations in need of a configuration and readiness step.
- Standardize method naming for getting a connection string from a container, where a popular client library expects a standardized URL format for connecting to the service provided by the container (E.g. a database URL for DB containers)
- Remove `DbContainer` class, as the two points above will render it redundant.

### DIFF
--- a/modules/mssql/testcontainers/mssql/__init__.py
+++ b/modules/mssql/testcontainers/mssql/__init__.py
@@ -1,11 +1,22 @@
 from os import environ
 from typing import Optional
 
-from testcontainers.core.generic import DbContainer
-from testcontainers.core.utils import raise_for_deprecated_parameter
+from typing_extensions import override
+
+from testcontainers.core.container import DockerContainer
+from testcontainers.core.utils import create_connection_string, raise_for_deprecated_parameter
+from testcontainers.core.waiting_utils import wait_container_is_ready
+
+ADDITIONAL_TRANSIENT_ERRORS = []
+try:
+    from sqlalchemy.exc import DBAPIError
+
+    ADDITIONAL_TRANSIENT_ERRORS.append(DBAPIError)
+except ImportError:
+    pass
 
 
-class SqlServerContainer(DbContainer):
+class SqlServerContainer(DockerContainer):
     """
     Microsoft SQL Server database container.
 
@@ -16,7 +27,7 @@ class SqlServerContainer(DbContainer):
             >>> import sqlalchemy
             >>> from testcontainers.mssql import SqlServerContainer
 
-            >>> with SqlServerContainer() as mssql:
+            >>> with SqlServerContainer("mcr.microsoft.com/azure-sql-edge:1.0.7") as mssql:
             ...    engine = sqlalchemy.create_engine(mssql.get_connection_url())
             ...    with engine.begin() as connection:
             ...        result = connection.execute(sqlalchemy.text("select @@VERSION"))
@@ -43,6 +54,17 @@ class SqlServerContainer(DbContainer):
         self.dbname = dbname
         self.dialect = dialect
 
+    @wait_container_is_ready(*ADDITIONAL_TRANSIENT_ERRORS)
+    def _wait_until_ready(self):
+        import sqlalchemy
+
+        engine = sqlalchemy.create_engine(self.get_connection_url())
+        try:
+            engine.connect()
+        finally:
+            engine.dispose()
+
+    @override
     def _configure(self) -> None:
         self.with_env("SA_PASSWORD", self.password)
         self.with_env("SQLSERVER_USER", self.username)
@@ -50,6 +72,11 @@ class SqlServerContainer(DbContainer):
         self.with_env("ACCEPT_EULA", "Y")
 
     def get_connection_url(self) -> str:
-        return super()._create_connection_url(
-            dialect=self.dialect, username=self.username, password=self.password, dbname=self.dbname, port=self.port
+        return create_connection_string(
+            dialect=self.dialect,
+            username=self.username,
+            password=self.password,
+            host=self.get_container_host_ip(),
+            port=self.get_exposed_port(self.port),
+            dbname=self.dbname,
         )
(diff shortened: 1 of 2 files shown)


santi:
Referencing https://github.com/testcontainers/testcontainers-python/issues/459 for context

alexanderankin:
would this have the potential to break existing code that depends on conventions from v3 and v4?

core/testcontainers/core/utils.py
```diff
@@ -53,6 +54,43 @@ def inside_container() -> bool:
...
+    username: Optional[str] = None,
+    password: Optional[str] = None,
+    driver: Optional[str] = None,
+    dbname: Optional[str] = None,
+) -> str:
+    """
+    Returns a connection URL following the RFC-1738 format.
+    Compatible with database clients such as SQLAlchemy and other popular database client libraries.
+
+    Example: postgres+psycopg2://myuser:mypassword@localhost:5432/mytestdb
```
jankatins, line 70: It would be nice to make "dialect: str | None " and document that setting it to None or "" will remove the dialect from the URL (psycopg v3 does need a url without dialect!)

core/testcontainers/core/utils.py
```diff
@@ -53,6 +54,43 @@ def inside_container() -> bool:
     return os.path.exists("/.dockerenv")
 
 
+def create_connection_string(
```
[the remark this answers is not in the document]
jankatins: Ok, saw that it is also used for http -> closing this comment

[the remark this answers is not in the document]
jankatins: create_connection_url -> e.g. you couldn't create a kafka one with this, couldn't you?

core/testcontainers/core/utils.py
```diff
@@ -53,6 +54,43 @@ def inside_container() -> bool:
...
+    dialect: str,
```
jankatins, line 58: If this would be DB API only, I would say "dialect" is fine, but if this should also be used to generate http URLs, I would call this base_scheme.

core/testcontainers/core/utils.py
jankatins: I would actually split this into a generic "create_connection_url" (`scheme` instead of `dialect` and `driver`, `path` instead of `dbname`) and one call (`create_db_api_connection_url`?)  which takes db api inputs (like the current one) and calls `create_connection_url` with the corresponding arguments.

modules/azurite/testcontainers/azurite/__init__.py
```diff
@@ -33,10 +33,9 @@ class AzuriteContainer(DockerContainer):
             >>> from testcontainers.azurite import AzuriteContainer
             >>> from azure.storage.blob import BlobServiceClient
 
-            >>> with AzuriteContainer() as azurite_container:
-            ...   connection_string = azurite_container.get_connection_string()
+            >>> with AzuriteContainer("mcr.microsoft.com/azure-storage/azurite:3.29.0") as azurite_container:
             ...   client = BlobServiceClient.from_connection_string(
-            ...        connection_string,
+            ...        azurite_container.get_connection_string(),
```
jankatins, line 38: For `ArangoDbContainer` is get_conenction_url`, here it's '*_string'. Should we unify that?

modules/azurite/testcontainers/azurite/__init__.py
```diff
@@ -102,12 +101,7 @@ def get_connection_string(self) -> str:
...
-    def start(self) -> "AzuriteContainer":
-        super().start()
-        self._connect()
-        return self
-
     @wait_container_is_ready(OSError)
-    def _connect(self) -> None:
+    def _wait_until_ready(self) -> None:
         with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
             s.connect((self.get_container_host_ip(), int(self.get_exposed_port(next(iter(self.ports))))))
```
jankatins, line 107: Might it make sense to also pull this into utils as a `wait_for_port`? At least I had the need in my code to basically write something similar:

```py
        @wait_container_is_ready(NoSuchPortExposed)
        def _wait_for_port() -> None:
            zks.get_service_port("schemaregistry", 8085)

        _wait_for_port()
```

modules/influxdb/testcontainers/influxdb.py
```diff
@@ -60,40 +60,10 @@ def __init__(
...
-        See its documentation for the max number of retries or the timeout.
-        """
-
-        url = self.get_url()
-        response = get(f"{url}/health", timeout=1)
-        response.raise_for_status()
-
-        return response.json()
-
-    def get_influxdb_version(self) -> str:
```
jankatins, line 85: Should this function be kept? It's seems it has value outside the normal startup procedure?

modules/influxdb/tests/test_influxdb.py
```diff
@@ -18,39 +18,26 @@
...
+        ("influxdb:2.7.5", InfluxDb2Container, "v2.7.5"),
     ],
 )
 def test_influxdbcontainer_get_influxdb_version(
     image: str, influxdb_container_class: Type[InfluxDbContainer], expected_version: str
 ):
     with influxdb_container_class(image) as influxdb_container:
-        assert influxdb_container.get_influxdb_version().startswith(expected_version)
+        version = get(f"{influxdb_container.get_url()}/health").json()["version"]
+        assert version == expected_version
```
jankatins, line 40: This test implies that `influxdb_container.get_influxdb_version()` is API?

alexanderankin:
if i were to try to merge this, what should i keep in mind @santi ? Anything not already reflected here? I did a first pass and seems like i will have to revisit this several times in order to make sense of this fully...

also, like i asked above, is this breaking at all?

santi:
@alexanderankin Hey, sorry for the delay on this one. This is just a WIP draft that is not ready for a merge quite yet, and has probably drifted a bit from the `main` branch. I intend on getting this one up-to-date this week, and do a nice write-up on the implications and document the intention behind the changes / API standardisation it introduces.

PR #529, opened 2024-04-05
chore(codestyle): switch to `ruff` from `black` for code formatting

max-pfeiffer:
Changed pre-commit config. Some files
became re-formatted.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,18 +9,14 @@ repos:
       - id: trailing-whitespace
       - id: end-of-file-fixer
 
-  - repo: https://github.com/psf/black-pre-commit-mirror
-    rev: '24.2.0'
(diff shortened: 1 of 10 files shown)


santi (approved):
Love it! Always nice with fewer dependencies and tools for development setups. 🎉 

I am all for merging this, but I think @totallyzen also should have a look as he is the one that set this up in the first place.

alexanderankin (approved):

totallyzen (approved):
LGTM!

I changed the title to `chore(): ...` because this only impacts readability and I don't think we need it on the changelogs

totallyzen:
> he is the one that set this up in the first place

I like writing clean code, but sometimes copy/pasta is the way to go for productivity so I'm not immune either. :joy:

PR #531, opened 2024-04-08
refactor: make config monkeypatchable

CarliJoy:
I am using testcontainers within a library that provides some pytest-fixtures.
In order for this to work I have change some settings.

As I can not guarantee that that my lib is imported before testcontainers I need to monkeypatch the settings.
This is much easier if I only need to monkeypatch the config file and not all modules that use configurations.

I would argue that for a potential library as this, this is a better design.

Also one can easier see that the given UPERCASE variable is not a constant but rather a setting.

### DIFF
--- a/core/testcontainers/core/container.py
+++ b/core/testcontainers/core/container.py
@@ -6,13 +6,7 @@ from typing import TYPE_CHECKING, Optional
 import docker.errors
 from typing_extensions import Self
 
-from testcontainers.core.config import (
-    RYUK_DISABLED,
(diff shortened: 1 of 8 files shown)


alexanderankin:
what if we gave you the red carpet treatment and actually made an API for this, like

https://github.com/testcontainers/testcontainers-python/compare/editable_config?expand=1

alexanderankin:
thanks for the idea - i gave you credit in my merge commit in #532, going to close this for now

PR #519, opened 2024-04-01
Add common testcontainers labels

bearrito:
- Closes: https://github.com/testcontainers/testcontainers-python/issues/510

Aligns with other test container projects

Example: I also contribute to go project, those labels look like

```
  "maintainer": "docker@couchbase.com",
  "org.opencontainers.image.ref.name": "ubuntu",
  "org.opencontainers.image.version": "20.04",
  "org.testcontainers": "true",
  "org.testcontainers.lang": "go",
  "org.testcontainers.sessionId": "e01aa90cfb75a53fbd53776b8c2eb84a99e3f1c8a7103512468cf75735421176",
  "org.testcontainers.version": "0.30.0"
```

Java appears to do similar - https://github.com/testcontainers/testcontainers-java/blob/main/core/src/main/java/org/testcontainers/DockerClientFactory.java#L51

I didn't add in the image info as there wasn't an obvious way to get a handle on that nor obvious value. 
Another thing is that the python prefers `session-id` to `sessionId`. I'm not sure if there are any cross-language reasons to have those be identical, I left it alone.

Also this adds in tests for the label code.

PR #532, opened 2024-04-08
fix(core): make config editable to avoid monkeypatching.1

alexanderankin:
see #531


alexanderankin:
ok, i checked against #519 and shouldnt be too hard to rebase (if even necessary), unfortunately re #525 and #527 - 527 is a draft and 525 is a feat, which deprioritizes over things which can be considered fixes - unless the changes are inspired by tc-go or tc-java or other sources of truth on how to deal with things like mssql, may even be "breaking" even though they are "fixing" - i unfortunately have a lot of pent up sympathy for people doing unusual software things. nothing against the PR, we will figure it out and hopefully release it as a fix rather than a feat)

so with that said, not sure why i shouldnt merge this

alexanderankin:
The data class is not frozen

Issue #541, opened 2024-04-16
Feature: Seed data for DbContainer


OverkillGuy:
**What are you trying to do?**

In order to do meaningful tests using my fresh database, I wish to load arbitrary SQL scripts from my DbContainer-s.

I can do (and have done) this per-project, in test folders, but believe there's value in allowing an arbitrary list of SQL scripts to be run before the testcontainer is yielded, and making this a pytest fixture to reuse.

I do this usually by first passing a path to a folder with scripts, volume-mount it to `/seeds` path in the database container, and also pass a list of script files in that folder to be executed. Then before yielding the ready db-container, I loop over each `script` filename and do `container.exec_run(["mysql", "-e" f"source /seeds/{script}")`.

The `list` of scripts means we can feed both schema first, then arbitrary sample data files.

I suggest a new `DbContainer._seed()` method, defaulting to `raise NotImplementedError`, to be overriden per database, allowing a new optional parameter `seed`.

Each database then just has to define their `_seed()` function to the specific way to `exec_run` the scripts (`psql` for Postgres, `mysql` for MySQL CLI...)

**Why should it be done this way?**

I believe most users of testcontainers, beyond just testing connectivity with a database, want a simple-to-reproduce test environment with realistic datastructure and mock data. This is currently annoying to do locally, and can be upstreamed into a simple-looking `seed` parameter, which could enable the feature for most databases, by pushing it into `DbContainer` itself.

alexanderankin:
how does this compare to this functionality in the [mysql][my] and [postgres][pg] images?

[my]: https://github.com/docker-library/docs/tree/master/mysql#initializing-a-fresh-instance
[pg]: https://github.com/docker-library/docs/tree/master/postgres#initialization-scripts

OverkillGuy:
> how does this compare to this functionality in the [mysql](https://github.com/docker-library/docs/tree/master/mysql#initializing-a-fresh-instance) and [postgres](https://github.com/docker-library/docs/tree/master/postgres#initialization-scripts) images?

Good shout, I didn't know these existed (there goes my evening, reading the manual like I _should_ have done first thing)

Let's say, at best, this proposed `seed` option can get reduced to just the local path, and we mount it to those standard `/docker-entrypoint-initdb.d/` folders, simplifying most of the implementation
This can harmonize the feature for images that support it?

Or we could just scrap the feature request altogether because it exists already as part of most images, haha. Either way works for me, the feature gap itself is solved by the links provided.

[...]

PR #542, opened 2024-04-16
feat: Run seed scripts in DbContainer

OverkillGuy:
New capability of "seeding" a db container, by running a database-specific command or twelve to inject data (DB schema or sample data).

Ref #541, implemented for MySQL but generalizeable

### DIFF
--- a/core/testcontainers/core/generic.py
+++ b/core/testcontainers/core/generic.py
@@ -10,6 +10,7 @@
 #    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 #    License for the specific language governing permissions and limitations
 #    under the License.
+from pathlib import Path
 from typing import Optional
 
 from testcontainers.core.container import DockerContainer
@@ -69,7 +70,20 @@ class DbContainer(DockerContainer):
         self._configure()
         super().start()
         self._connect()
+        self._seed()
         return self
 
     def _configure(self) -> None:
         raise NotImplementedError
+
+    def _seed(self) -> None:
+        raise NotImplementedError
+
+    def _setup_seed(self, seed):
+        self.seed_scripts = None
+        self.seed_mount = "/seeds"  # TODO: Should it be configurable?
+        if seed is not None:
+            seed_localpath, seed_scripts = seed
+            seed_abspath = Path(seed_localpath).absolute()
+            self.with_volume_mapping(seed_abspath, self.seed_mount)
+            self.seed_scripts = seed_scripts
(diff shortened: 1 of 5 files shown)


modules/mysql/testcontainers/mysql/__init__.py
```diff
@@ -48,12 +67,14 @@ def __init__(
...
         dbname: Optional[str] = None,
         port: int = 3306,
+        seed: Optional[tuple[str, list[str]]] = None,
         **kwargs,
     ) -> None:
         raise_for_deprecated_parameter(kwargs, "MYSQL_USER", "username")
         raise_for_deprecated_parameter(kwargs, "MYSQL_ROOT_PASSWORD", "root_password")
         raise_for_deprecated_parameter(kwargs, "MYSQL_PASSWORD", "password")
         raise_for_deprecated_parameter(kwargs, "MYSQL_DATABASE", "dbname")
+
```
OverkillGuy, line 77:
```suggestion
```

core/testcontainers/core/generic.py
```diff
@@ -69,7 +70,20 @@ def start(self) -> "DbContainer":
...
         super().start()
         self._connect()
+        self._seed()
         return self
 
     def _configure(self) -> None:
         raise NotImplementedError
+
+    def _seed(self) -> None:
+        raise NotImplementedError
```
OverkillGuy, line 80: Just realized we of course shouldn't penalize DbContainer variants for not defining this method, should do something like `if not self.seeds: return` and only otherwise raise this error

modules/mysql/testcontainers/mysql/__init__.py
```diff
@@ -40,6 +40,25 @@ class MySqlContainer(DbContainer):
...
+        is perfect for schema and sample data. The format is a tuple, made up of the
+        path to local data, and list of script files to load. Each script will be loaded
+        by exec in the container, using the MySQL root user, before yielding the
+        container. Any errors loading the scripts, will cause a ValueError.
+
+        .. doctest::
+
+            >>> import sqlalchemy
+            >>> from testcontainers.mysql import MySqlContainer
+            >>> seed_data = ("../../tests/", ["schema.sql", "data.sql"])
```
OverkillGuy, line 54: Tradeoff: the docstring is ugly due to weird relative path to real sql files, but if I write what I mean:
```suggestion
            >>> seed_data = ("db/scripts/", ["schema.sql", "data.sql"])
```

Well then the doctests fail, which won't work either!

alexanderankin:
see discussion in #541

PR #552, opened 2024-04-21
fix(mysql): Add seed support in MySQL

OverkillGuy:
Ref #541.
New capability of "seeding" a db container using image's support for /docker-entrypoint-initdb.d/ folder.

Using the "transferable" system, borrowed from Kafka. 

Updates DbContainer to have a new (NOOP-default) `_transfer_seed()` method, run after `_start()` and before `_connect()`, to allow the folder transfer.

Currently implemented only in MySQL, but extensible to others that use the `/docker-entrypoint-initdb.d/` system.

### DIFF
--- a/core/testcontainers/core/generic.py
+++ b/core/testcontainers/core/generic.py
@@ -71,7 +71,11 @@ class DbContainer(DockerContainer):
         self._configure()
         super().start()
         self._connect()
+        self._transfer_seed()
         return self
 
     def _configure(self) -> None:
         raise NotImplementedError
+
+    def _transfer_seed(self) -> None:
+        pass
--- a/core/testcontainers/core/generic.py
+++ b/core/testcontainers/core/generic.py
@@ -70,8 +70,8 @@ class DbContainer(DockerContainer):
     def start(self) -> "DbContainer":
         self._configure()
         super().start()
-        self._connect()
         self._transfer_seed()
+        self._connect()
         return self
 
     def _configure(self) -> None:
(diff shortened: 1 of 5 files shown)


core/testcontainers/core/generic.py
```diff
@@ -70,8 +70,12 @@ def _create_connection_url(
...
         self._configure()
         super().start()
+        self._transfer_seed()
         self._connect()
         return self
 
     def _configure(self) -> None:
         raise NotImplementedError
+
+    def _transfer_seed(self) -> None:
```
OverkillGuy, line 80: To be discussed if this is a universal-enough feature to warrant this in DbContainer, as opposed to supporting it on a per-db basis.

OverkillGuy:
Updated to use "transferable" like kafka does, which makes this support remote docker engines!

modules/mysql/testcontainers/mysql/__init__.py
```diff
@@ -86,3 +107,14 @@ def get_connection_url(self) -> str:
...
             dialect="mysql+pymysql", username=self.username, password=self.password, dbname=self.dbname, port=self.port
         )
+
+    def _transfer_seed(self) -> None:
+        if self.seed is None:
+            return
+        src_path = Path(self.seed)
+        dest_path = "/docker-entrypoint-initdb.d/"
+        with BytesIO() as archive, tarfile.TarFile(fileobj=archive, mode="w") as tar:
+            for filename in src_path.iterdir():
```
OverkillGuy, line 117: Note the classic `src_path.walk()` is actually not available until Python 3.12 for PosixPath (linux).
This lib seems to support lower versions, so `Path.iterdir()` is a good alternative

PR #556, opened 2024-04-24
fix(core): add empty _configure to DockerContainer

christianaaronschroeder:
I've used DbContainer in the past as a parent to a class where I redefined `_configure()` and did not redefine `start()`. I'm now using DockerContainer and wanted to follow the same pattern but it doesn't have a `_configure()` function called in `start()`. I figure this gives the easy option without being invasive?

tests pass


alexanderankin:
ok, its in the java version - https://github.com/testcontainers/testcontainers-java/blob/18ea976a640e43e3fa1a1da7285c4630fd9cdbac/core/src/main/java/org/testcontainers/containers/GenericContainer.java#L701 - , i say why not

Issue #557, opened 2024-04-26
Bug: Wrong return type hint for exec of DockerContainer


levrik:
**Describe the bug**

The return type of `exec` of `DockerContainer` is defined as `tuple[int, str]` while it is `tuple[int, bytes]` at runtime.

PR #561, opened 2024-04-29
fix: added types to exec & tc_properties_get_tc_host

Dandiggas:
https://github.com/testcontainers/testcontainers-python/issues/557 -  trying to solve this issue by adding types.

### DIFF
--- a/core/testcontainers/core/config.py
+++ b/core/testcontainers/core/config.py
@@ -47,7 +47,7 @@ class TestcontainersConfiguration:
     ryuk_reconnection_timeout: str = RYUK_RECONNECTION_TIMEOUT
     tc_properties: dict[str, str] = field(default_factory=read_tc_properties)
 
-    def tc_properties_get_tc_host(self):
+    def tc_properties_get_tc_host(self) -> str | None:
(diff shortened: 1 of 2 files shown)


errordeveloper (approved):
this LGTM overall, thanks @Dandiggas!

errordeveloper:
@kiview could you please approve the workflows to run?

alexanderankin:
We'll take a look

alexanderankin:
yep so we support python 3.9 for now so cant use the pipe syntax for `typing.Union` - but the other type is indeed wrong so we could merge a pr that just fixes a single type, would probably prefer a PR that fixes types more comprehensively like #504 - if you want to provide feedback on that branch, that could help it along - it was only not merged because i was not confident in 100% of the choices made in there at the time.

ill leave this open in case you want to replace with `from typing import Union`

Dandiggas:
@alexanderankin Now supports python 3.9, please review when possible

alexanderankin:
thank you for the contribution! if you are interested in improving typing in this library are other types that I have started but did not feel confident enough to merge #504

PR #565, opened 2024-05-08
fix(keycloak): add realm imports

stefnnn:
This PR adds an option to start a keycloak container with importing one or more realms. This mirrors a feature present in the java keycloak testcontainer.

### DIFF
--- a/modules/keycloak/testcontainers/keycloak/__init__.py
+++ b/modules/keycloak/testcontainers/keycloak/__init__.py
@@ -65,7 +65,7 @@ class KeycloakContainer(DockerContainer):
         if self.has_realm_imports:
             self.cmd += " --import-realm"
         self.with_command(self.cmd)
- 
+
     def get_url(self) -> str:
         host = self.get_container_host_ip()
         port = self.get_exposed_port(self.port)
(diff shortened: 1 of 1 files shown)


alexanderankin:
if you can make it backwards compatible (ie not break tests) then i dont see why we wouldnt merge

stefnnn:
My bad! I have fixed a bug I introduced, that prevented to wait for the first user to be added. Additionally I made the cmd configurable, which I think make sense, depending on the use caes.

However, I cannot get the tests running locally, neither on this branch nor on main, so I have a hard time verify whether it works consitently now.

alexanderankin:
so it seems straightforward enough, so lgtm!

also idk if you accounted for this but the command can be a list or a string. since backwards (or anything beyond basic) compatibility is really not a top concern for community modules, im going to merge this, but feel free to follow up before we do a release. not sure how soon that will be since im pretty backlogged and i dont think we have anything to release that is super critical (happy to be corrected about this).

stefnnn:
Backward compatibility will not be an issue, as there was no way to set the docker command in the previous version - it was automatically overwritten to _DEFAULT_DEV_COMMAND by the start() function. So I guess it's good enough for now. Thanks a lot for the thourough review and time!

Issue #562, opened 2024-05-06
Feature: Docker private registry support


claussa:
**What are you trying to do?**

I want to use testcontainers with an image store on a private registry. Correct me if I am wrong, It is currently not possible on testcontainers-python but it is possible on testcontainers-java.

**Why should it be done this way?**

Example on how it is done on testcontainers-java: https://github.com/testcontainers/testcontainers-java/blob/994b385761dde7d832ab7b6c10bc62747fe4b340/core/src/main/java/org/testcontainers/utility/RegistryAuthLocator.java#L37

alexanderankin:
thank you for the example - before reading that code, i never understood what this meant - i suppose in ci environments you have to get creative on how to pull your images, so that makes sense. if i understand correctly, this is just passing an auth parameter for pulling an image if it doesnt exist (and workaround is then just to pull the image ahead of time?)

Tranquility2:
1. Will this need support for more than one auth server? or basic support is for a single config?
the basic example is :
```
{
    "auths": {
        "registry.example.com:5000": {
            "auth": "bXlfdXNlcm5hbWU6bXlfcGFzc3dvcmQ="
        }
    }
}
```
note the plural `auths` :) 

2. I assume this does not include `~/.docker/config.json` as its another feat all by itself right?

Tranquility2:
Good news. I think I got it working for a single registry:
```
 echo "$DOCKER_AUTH_CONFIG"
{
        "auths": {
                "localhost:5000": {
                        "auth": "..."
                }
        },
}

>>> from testcontainers.redis import RedisContainer
>>> with RedisContainer() as redis_container:
...     redis_client = redis_container.get_client()
...
Pulling image testcontainers/ryuk:0.7.0
Container started: bb76132ce865
Waiting for container <Container: bb76132ce865> with image testcontainers/ryuk:0.7.0 to be ready ...
Pulling image localhost:5000/redis:latest
Container started: a48fd2609171
Waiting for container <Container: a48fd2609171> with image localhost:5000/redis:latest to be ready ...
Waiting for container <Container: a48fd2609171> with image localhost:5000/redis:latest to be ready ...
Waiting for container <Container: a48fd2609171> with image localhost:5000/redis:latest to be ready ...
```

alexanderankin:
is there a PR to look at or is that just bulit-in functionality from docker-py?

Tranquility2:
[...]

PR #566, opened 2024-05-08
feat(core): Private registry

Tranquility2:
Ref #562

This enhancement adds capability to utilize the env var `DOCKER_AUTH_CONFIG` in-order to login to a private docker registry.


alexanderankin:
i wonder if this is what the java client does underneath, have almost no opportunity to find out independently. i explored the testcontainers side of java yesterday and realized that i need to look at docker-java to find any real answers anyways.

Tranquility2:
Added tests

Tranquility2:
Some documentation on how I tested it:
https://gist.github.com/Tranquility2/20652b77a012ee54ae1c8ca021aea1ff

alexanderankin:
wdyt about testing using this module - https://github.com/testcontainers/testcontainers-python/tree/main/modules/registry - i know it inverts the dependencies on their head a bit. maybe we can just copy and paste the code as it is just a part of the test suite and therefore not actually being published to pypi. doesn't have to be a part of this PR, just think that would be the most comprehensive/on-brand test.

maybe its not actually a blocker, just pausing to think cause when this is merged, there is no more fixes for 4.4 line (RP will set the version in #551 to `4.5.x`)

Tranquility2:
was thinking the same thing :) 
I had the exact same doubt as its not really a best practice but fits like a glove for a more comprehensive test.
Do you think it should be in this PR or another? In any case I'll be happy to add it.

alexanderankin:
Ok, so I'll just compile some research here for the future:

* the [docker api expects a json object](https://docs.docker.com/engine/api/v1.45/#section/Authentication) with `username` and `password` properties - base 64 encoded in the `X-Registry-Auth` header
* the [login method](https://github.com/docker/docker-py/blob/a3652028b1ead708bd9191efb286f909ba6c2a49/docker/api/daemon.py#L97-L100) will [call `add_auth`](https://github.com/docker/docker-py/blob/a3652028b1ead708bd9191efb286f909ba6c2a49/docker/api/daemon.py#L151) - which [saves things into a dict](https://github.com/docker/docker-py/blob/a3652028b1ead708bd9191efb286f909ba6c2a49/docker/auth.py#L305-L307) - which get used to set the auth header during [inspect](https://github.com/docker/docker-py/blob/a3652028b1ead708bd9191efb286f909ba6c2a49/docker/api/image.py#L276-L283), [pull](https://github.com/docker/docker-py/blob/a3652028b1ead708bd9191efb286f909ba6c2a49/docker/api/image.py#L407-L415) and [push](https://github.com/docker/docker-py/blob/a3652028b1ead708bd9191efb286f909ba6c2a49/docker/api/image.py#L481-L489) (and also in [plugin](https://github.com/docker/docker-py/blob/a3652028b1ead708bd9191efb286f909ba6c2a49/docker/api/plugin.py) and [service](https://github.com/docker/docker-py/blob/a3652028b1ead708bd9191efb286f909ba6c2a49/docker/api/service.py)) but not create container, like java does: [createCmdExec](https://github.com/docker-java/docker-java/blob/b29527510d9f7c16b30b138ae4a2f6703d88bb01/docker-java-core/src/main/java/com/github/dockerjava/core/exec/CreateContainerCmdExec.java#L35-L37), [AbstractDockerCommandExec](https://github.com/docker-java/docker-java/blob/b29527510d9f7c16b30b138ae4a2f6703d88bb01/docker-java-core/src/main/java/com/github/dockerjava/core/exec/AbstrDockerCmdExec.java#L78-L85).

If this actually works and doesn't break anything, I think this is as close as were going to get.

alexanderankin:
> Do you think it should be in this PR or another? In any case I'll be happy to add it.

probably another at this point. its more than welcome, but have to keep merging things or else they get stale and conflicted and unmergable.

Tranquility2:
@alexanderankin sorry to say but the latest changes broke the new planned test. 
Now the config is loaded even before the `DockerRegistryContainer` invocation/build so the test cannot build DOCKER_AUTH_CONFIG using its `registry_url` (which is dynamic..)

Even if I do manage to set it before starting `DockerRegistryContainer` it will fail as it doesn't have the registry at this point.
I can do something like:
```
def get_docker_auth_config() -> Optional[str]:
    return c.docker_auth_config or os.getenv("DOCKER_AUTH_CONFIG")
```
That solves this race but makes all the config part that you added a bit redundant.

In any case we need a way to update this after the the first `container` init `testcontainers_config` right? got any ideas?

alexanderankin:
> got any ideas?

Yes i can try to take a look but I think seeing your code (even with that change) would help me understand, I don't think I'm following based on your description. It worked last time for the contribution, then some minor changes from me during review, so hopefully it will work again.

Tranquility2:
https://github.com/testcontainers/testcontainers-python/pull/582

Issue #570, opened 2024-05-12
Bug: Docker Compose test is broken


Tranquility2:
**Describe the bug**

It seems that with the latest version of docker compose installed the `test_compose_logs` fails

**To Reproduce**

1. Update docker (Linux machine) 
*My guess is `docker-compose-plugin` 
2. Pull main branch of `testcontainers-python`
3. Run the tests
Can be as simple as
```bash
pytest core/tests/test_compose.py
```

** Example **
```bash
=================================================== test session starts ====================================================
platform linux -- Python 3.9.18, pytest-7.4.3, pluggy-1.4.0
rootdir: /home/roy/projects/testcontainers-python
configfile: pyproject.toml
plugins: anyio-4.3.0, asyncio-0.23.5, cov-4.1.0
asyncio: mode=strict
collected 11 items

core/tests/test_compose.py::test_compose_no_file_name PASSED                                                         [  9%]
core/tests/test_compose.py::test_compose_str_file_name PASSED                                                        [ 18%]
core/tests/test_compose.py::test_compose_list_file_name PASSED                                                       [ 27%]
core/tests/test_compose.py::test_compose_stop PASSED                                                                 [ 36%]
core/tests/test_compose.py::test_compose_start_stop PASSED                                                           [ 45%]
core/tests/test_compose.py::test_compose PASSED                                                                      [ 54%]
core/tests/test_compose.py::test_compose_logs FAILED                                                                 [ 63%]
core/tests/test_compose.py::test_compose_ports
------------------------------------------------------ live log call -------------------------------------------------------
INFO     testcontainers.core.waiting_utils:waiting_utils.py:54 Waiting for DockerCompose(context=PosixPath('/home/roy/projects/testcontainers-python/core/tests/compose_fixtures/port_single'), compose_file_name=None, pull=False, build=False, wait=True, env_file=None, services=None, docker_command_path=None) to be ready ...
PASSED                                                                                                               [ 72%]
core/tests/test_compose.py::test_compose_multiple_containers_and_ports PASSED                                        [ 81%]
core/tests/test_compose.py::test_exec_in_container
------------------------------------------------------ live log call -------------------------------------------------------
[...]

PR #571, opened 2024-05-12
fix(core): remove version from compose tests

Tranquility2:
This would address #570
Removing the version from the compose.yaml files used in tests.

Please see https://forums.docker.com/t/docker-compose-yml-version-is-obsolete/141313


alexanderankin:
Pleasure doing business

PR #585, opened 2024-05-27
feat(core): Image build (Dockerfile support)

Tranquility2:
As part of the effort described, detailed and presented on #559 
(Providing the implementation for #83 - Docker file support and more)
This is the first PR (out of 4) that should provide all the groundwork to support image build.

This would allow users to use custom images:
```python
  with DockerImage(path=".") as image:
            with DockerContainer(str(image)) as container:
                # Test something with/on custom image
```

Next in line is:
`feat(core): Added SrvContainer`
And later on:
`feat(core): Added FastAPI module`
`feat(core): Added AWS Lambda module`
(all of the above can be overviewed on #559)


juangon:
Would be great having this :+1:

PR #595, opened 2024-06-01
feat(core): Added ServerContainer

Tranquility2:
As part of the effort described, detailed and presented on #559 
This is the seconds PR (out of 4) that should provide all the groundwork to support containers running a server.

This would allow users to use custom images:
```python
with DockerImage(path=".", tag="test:latest") as image:
    with ServerContainer(port=9000, image=image) as srv:
        # Test something with/on the server using port 9000
```

Next in line are:
`feat(core): Added FastAPI module`
`feat(core): Added AWS Lambda module`

---
Based on the work done on #585
Expended from issue #83


alexanderankin:
some thoughts:
1. can you rename as ServerContainer
2. can we make this a module under modules? I guess if we do this we have to figure out how to make those modules depend on each other but assuming we can make that work, how do we feel about this.

Tranquility2:
Yes, you are right, `ServerContainer` is a much better name and it correlates to your second point, I do think its very general and more in par with `DBContainer` than a specific module (thats why I placed it under generics). I think the modules should remain specific product related. I'll try and support my claim with the java implementation of "nginx container" (which I'll probably add later on 😅 ) https://github.com/testcontainers/testcontainers-java/blob/6658a2c0a880d01c6d402ea9a4cb5f72eb15083c/modules/nginx/src/main/java/org/testcontainers/containers/NginxContainer.java#L12C13-L12C29
Notice it uses `GenericContainer` which is (in a way) the combination of `DBContainer` and `ServerContainer`.

So to conclude I think we should keep `ServerContainer` under generics, and maybe plan in the future to merge `DBContainer` and `ServerContainer` (?) into `GenericContainer` (but to be honest I think the separation is good at this stage)

alexanderankin:
for context, we are trying to remove DbContainer as well

Tranquility2:
I can offer another alternative, completely skip  `ServerContainer` and just make the modules repeat some code, I can also check it out and propose a draft if you like.

alexanderankin:
opened https://github.com/Tranquility2/testcontainers-python/pull/2

Tranquility2:
rebased with new improvements

Issue #568, opened 2024-05-10
Feature: add MosquittoContainer


f18m:
**What are you trying to do?**

I'm writing integration tests for a Python project (https://github.com/f18m/ha-alarm-raspy2mqtt) that publishes messages to an MQTT broker.
I would like to use testcontainers to create the integration tests and noticed that there is no specialization available for MQTT and for its most-used broker which is Mosquitto.
So I'd like to propose the addition of a MosquittoContainer class to testcontainers-python.

**Why should it be done this way?**

There are a number of helpers that could be provided out of the box to ease integration testing of MQTT applications.

I can provide a PR to add my own MosquittoContainer implementation if there is interest by the project maintainers...

alexanderankin:
sure, i cant promise it will be a priority but we've been merging a bunch of these. the hivemq seems to be king in the enterprise world with this so id say if we do merge a mqtt one it has to be this one:

https://java.testcontainers.org/modules/hivemq/

f18m:
hi @alexanderankin ,
thanks for the reply. Actually I've been developing the specialization of DockerContainer for the Mosquitto broker (https://mosquitto.org/) which is a very popular choice at least in home automation environment... I could contribute that if there's interest...

alexanderankin:
hm, i suppose we could have a module with multiple classes in it

f18m:
hi @alexanderankin ,
I opened PR #599 to propose my implementation of a MosquittoContainer class.
Several usage examples can be found at https://github.com/f18m/rpi2home-assistant/blob/main/tests/test_integration.py

PR #599, opened 2024-06-05
fix(mqtt): Add mqtt.MosquittoContainer (#568)

f18m:
This PR is adding a new MosquittoContainer class that helps creating integration tests for MQTT clients.
The MosquittoContainer class contains a bunch of methods to help with testing:
* checking number of messages received 
* watching topics
* check last payload published on a particular topic
* etc

This PR lacks tests. I can add them if there is interest in this PR...


modules/mqtt/testcontainers/mosquitto/__init__.py
```diff
@@ -0,0 +1,247 @@
...
+
+        interval = 1.0
+        timeout = 5
+        start = time.time()
+        while True:
+            duration = time.time() - start
+            if client.is_connected():
+                return
+            if duration > timeout:
+                raise TimeoutError(f"Failed to estabilish a connection after {timeout:.3f} " "seconds")
```
alexanderankin, line 96: this is a double wait - you're supposed to throw from wait_container_is_ready i think

modules/mqtt/testcontainers/mosquitto/__init__.py
```diff
@@ -0,0 +1,247 @@
...
+        def get_last_payload(self):
+            return self.last_payload
+
+        def get_rate(self):
+            duration = time.time() - self.timestamp_start_watch
+            if duration > 0:
+                return self.count / duration
+            return 0
+
+    def on_message(client: mqtt_client.Client, mosquitto_container: "MosquittoContainer", msg: mqtt_client.MQTTMessage):
```
alexanderankin, line 163: this is quite a tight coupling, probably too much for a testing utility

modules/mqtt/testcontainers/mosquitto/__init__.py
```diff
@@ -0,0 +1,247 @@
...
+        # wait till we get the first message from the topic;
+        # this wait will be up to 'sys_interval' second long (see mosquitto.conf)
+        try:
+            message = self.msg_queue.get(block=True, timeout=5)
+            return int(message.payload.decode())
+        except Queue.Empty:
+            return 0
+
+    def watch_topics(self, topics: list):
+        client, err = self.get_client()
```
alexanderankin, line 196: same here, seems to be doing too much

modules/mqtt/testcontainers/mosquitto/__init__.py
```diff
@@ -0,0 +1,247 @@
...
+        ret = self.client.publish(topic, payload)
+        ret.wait_for_publish(timeout=2)
+        if not ret.is_published():
+            raise RuntimeError(f"Could not publish a message on topic {topic} to Mosquitto broker: {ret}")
+
+    def print_logs(self) -> str:
+        print("** BROKER LOGS [STDOUT]:")
+        print(self.get_logs()[0].decode())
+        print("** BROKER LOGS [STDERR]:")
+        print(self.get_logs()[1].decode())
```
alexanderankin, line 247: this is not really necessary to add

alexanderankin (changes requested):
interest is there, would merge a more generic and simpler/smaller version

f18m:
> interest is there, would merge a more generic and simpler/smaller version

ok thanks for the review!
let me try to address the various points you raised

modules/mqtt/testcontainers/mosquitto/__init__.py
f18m: right -- good catch. I've gone through the code again and I better understood how it works...

modules/mqtt/testcontainers/mosquitto/__init__.py
f18m: I understand your point of view. OTOH the reason for using test-containers is to run integration tests, right?
To do integration testing for MQTT, I think it's a basic necessity to be able to inspect what the client has published, isn't it?
I mean: whoever is going to use this test-container will need to have some functionality like this one right?
Of course I'm open to change it to be more generic, if there's a way to do it 
(e.g. I can change the code so that there is no need to explicitly "watch" a topic -- MosquittoContainer might just be creating a WatchedTopicInfo instance for every possible topic where traffic is received)...
What's your thought on that?

modules/mqtt/testcontainers/mosquitto/__init__.py
f18m: right now this is an opt-in feature: if the user wants to "watch" a topic, he can use the watch_topics() API. If it's not used, there's no (performance) impact during the test...
so perhaps it's a "nice to have" that can stay?

modules/mqtt/testcontainers/mosquitto/__init__.py
f18m: removed

f18m:
@alexanderankin please note that I also checked in the "integration-test-mosquitto.conf" that is bind-mounted in the __init__ ctor.
I'm not sure what's the best way to deliver this kind of resources (the conf file) inside the project... so I just dropped it in the same folder of the __init__.py file, which is what the __init__ ctor uses for creating the bind-mount...

modules/mqtt/testcontainers/mosquitto/__init__.py
f18m: resolving the conversation as I think I put a better implementation in place

modules/mqtt/testcontainers/mosquitto/__init__.py
alexanderankin: I mean I am using this as my reference - https://github.com/testcontainers/testcontainers-java/blob/main/modules/hivemq/src/main/java/org/testcontainers/hivemq/HiveMQContainer.java - will have to review and reconsider

alexanderankin:
mounting files is probably the most straight forward way to do it but I would move it into the `_configure` (or `start` if you want ultimate backwards compatibility) method rather than the constructor. there is a concept of a transferable interface which can be implemented for strings and bytes/io but it is not merged, there is a couple commits here which should eventually make it into the project - https://github.com/testcontainers/testcontainers-python/compare/2bd20c6a9ab811343beee150a4816711848fb9ea...alexanderankin:testcontainers-python:TCP-393_transferable - the code in the kafka container which does the some thing is even more concise. i think there may be some chance that this transferable thing is going to make it into the project through one of the PR's that are doing database seeding (helpers around `init-db.d`)

modules/mqtt/testcontainers/mosquitto/__init__.py
alexanderankin: This sounds a bit conservative but I'm not sure I want to tightly couple folks' workflow with testcontainers - like you should still have very good logging in your application because hopefully, your application will make it to production, and then it will be up to cloudwatch/kibana/grafana to help you debug. if you were debugging this whole time with testcontainers utility methods, now you are going to be in a tight spot.

modules/mqtt/testcontainers/mosquitto/__init__.py
f18m: removed

f18m:
> ```python
> Path(__file__).parent
> ```

moved into start() the volume mapping bind!

f18m:
@alexanderankin just take a look again when you have time, thanks

alexanderankin:
this is probably what id like to merge, lmk if this works for you:

https://github.com/testcontainers/testcontainers-python/compare/feature/mosquitto

f18m:
> this is probably what id like to merge, lmk if this works for you:
> 
> https://github.com/testcontainers/testcontainers-python/compare/feature/mosquitto

super! I see you added support also for Mosquitto 1.x, which is something I never used.  I think it's good to have.

I think it's just good. Once merged I will rework my project to make use of the MosquittoContainer class and report back any possible issue. Thanks

alexanderankin:
if you want you can pip install into your project from the branch and do some testing before it is released as well

alexanderankin:
ok ill merge at some point today.

f18m:
> ok ill merge at some point today.

yeah I don't think I will manage to adapt my project that relies on MosquittoContainer fast enough. 
But next week I think I will have some time to do that and will report back

alexanderankin:
gonna merge because it modifies the poetry lock file and dont wanna have conflicts, still a chance to try it before it gets released

f18m:
hi @alexanderankin ,
I'm trying to understand how to test an unreleased version of testcontainers-python.
However I see that the CI/CD step 'Release Please / publish (push) ' is Skipped... 
are testcontainers-python artifacts published somewhere after every commit?

f18m:
> hi @alexanderankin , I'm trying to understand how to test an unreleased version of testcontainers-python. However I see that the CI/CD step 'Release Please / publish (push) ' is Skipped... are testcontainers-python artifacts published somewhere after every commit?

Never mind: I installed the main branch locally from git into my python path.

PR #612, opened 2024-06-21
feat(core): Added Generic module

Tranquility2:
As part of the effort described, detailed and presented on https://github.com/testcontainers/testcontainers-python/pull/559
This is the third PR (out of 4) that should provide all the groundwork to support containers running a server.

As discussed on #595 this PR aims to refactor the `ServerContainer` under a new dedicated module called "generic"
The idea is that this module could include multiple generic implementations such as ```server.py``` with the proper documentation and examples to allow users simpler usage and QOL.
        
Next in line is ```feat(core): Added AWS Lambda module```

Based on the work done on https://github.com/testcontainers/testcontainers-python/pull/585 and #595
Expended from issue https://github.com/testcontainers/testcontainers-python/issues/83

---
Please note an extra commit is included to simulate the relations when importing between and with other modules.


Tranquility2:
I'll remove the extra commit `Test module import' once we feel the PR is ready

alexanderankin:
I actually really love the test module import, if it can be `test_module_import` instead of just one word, or maybe a different name, this is really useful for the future as a proof that it is working. I think it should be enforced with CI.

I kind of want to think about this some more before merging, as its a fairly major change. but one that I think make sense.

i have to double check what we are doing with the httpx in the poetry files because afaik, you should not need to be adding it to the extras. we already depend on urllib3, we should just do the same to httpx, if we are doing to rely on it this heavily.

Tranquility2:
I'll be happy to add `test_module_import` but think it should go as a separate commit right?

Issue #610, opened 2024-06-20
Bug: DockerImage path arg is misleading


black-snow:
**Describe the bug**

`testcontainers.core.image.DockerImage` has a constructor arg `path` with the following docs:

> path – Path to the Dockerfile to build the image

But actually this doesn't seem to be true. It's rather the context, the path to the directory to build in and the dockerfile itself must be named `Dockerfile`. Especially for testing this is very unhandy. I often have several dockerfiles in the same directory. I'd expect bein able to pass a `context` and the _actual_ `path` to the dockerfile, like

```python
with DockerImage(dockerfile="./caseA.Dockerfile", context=".", tag="casea:latest") as image:
     pass
```

**To Reproduce**

* have a `test.Dockerfile`
* `with DockerImage(path="./test.Dockerfile", tag="test") as image: ...`
* see not too helpful error message from docker itself bubble up

**Runtime environment**

not relevant

---
P.S.: build-args would also be handy

alexanderankin:
so, that's a mistake - I'm pretty busy with my job these days but will review patches, seems like something to fix (and add re: build/args).

i might even argue for calling adding build/args a fix, as it makes our api more congruent with the docker api.

black-snow:
I'll take a look at it.

black-snow:
Err... `make tests` in the devcontainer leaves me with 32 failed, 42 passed, 3 warnings in 61.63s

alexanderankin:
i think that makes sense if you are on windows. its on my list to find a piece of terraform code that creates a windows VM in aws, so i can develop against windows, lots of challenges there. (cc @kiview)

alexanderankin:
@black-snow you should be `poetry install --with dev --all-extras` if you are trying to run the full test suite (like `pytest -s`)

black-snow:
@alexanderankin I'm on Ubuntu 24 right now. Even with all the extras 30 tests still fail.

black-snow:
From https://docker-py.readthedocs.io/en/stable/images.html#docker.models.images.ImageCollection.build it's not very clear what exactly the `path` is either. Looking at the [code](https://github.com/docker/docker-py/blob/main/docker/api/build.py#L355) I'm pretty sure this is not what I expect. So I might have to redirect this to docker-py first.

It's totally fine to `docker build -t whatevs -f some/sub/dir/my.Dockerfile .` - yet the code seems to require the dockerfile to be top-level in the `path`, which is treated as context.

[...]

PR #614, opened 2024-06-21
feat(core): add buildargs to DockerImage

black-snow:
Adds `--build-arg` equivalent as mentioned in https://github.com/testcontainers/testcontainers-python/issues/610

I don't get the tests to pass so this is actually untested!

Also, I'd rather move it into a separate, well-named test instead of stuffing everything in there but I didn't want to refactor so much. I'll perhaps refactor in another PR.

### DIFF
--- a/core/testcontainers/core/image.py
+++ b/core/testcontainers/core/image.py
@@ -1,5 +1,5 @@
 from os import PathLike
-from typing import TYPE_CHECKING, Optional, Union
+from typing import TYPE_CHECKING, Optional, Union, Dict
 
 from typing_extensions import Self
 
@@ -25,6 +25,7 @@ class DockerImage:
 
     :param tag: Tag for the image to be built (default: None)
     :param path: Path to the Dockerfile to build the image
+    :param build_args: Dict of build args; equivalent to CLI's --build-arg
     """
 
     def __init__(
@@ -33,6 +34,7 @@ class DockerImage:
         docker_client_kw: Optional[dict] = None,
         tag: Optional[str] = None,
         clean_up: bool = True,
+        build_args: Dict[str, str] = None,
         **kwargs,
     ) -> None:
         self.tag = tag
@@ -42,11 +44,14 @@ class DockerImage:
         self._kwargs = kwargs
         self._image = None
         self._logs = None
+        self._build_args = build_args
 
     def build(self, **kwargs) -> Self:
         logger.info(f"Building image from {self.path}")
         docker_client = self.get_docker_client()
-        self._image, self._logs = docker_client.build(path=str(self.path), tag=self.tag, **kwargs)
+        self._image, self._logs = docker_client.build(
+            path=str(self.path), tag=self.tag, buildargs=self._build_args, **kwargs
+        )
         logger.info(f"Built image {self.short_id} with tag {self.tag}")
         return self
 
--- a/core/tests/test_core.py
+++ b/core/tests/test_core.py
@@ -45,8 +45,8 @@ def test_docker_image(test_image_tag: Optional[str], test_cleanup: bool, test_bu
     with tempfile.TemporaryDirectory() as temp_directory:
         # It's important to use a random string to avoid image caching
         random_string = "Hello from Docker Image! " + str(random.randint(0, 1000))
-        build_arg_name = test_build_arg[0]
-        build_arg_value = test_build_arg[0]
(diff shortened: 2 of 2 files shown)


Tranquility2:
Hi @black-snow, if I can offer some assistance, it looks like the number of steps just changed 
image
probaly from adding
```
                ARG {build_arg_name}
                ENV {build_arg_name}=${build_arg_name}
```
I take the blame for the original test it was way too hardcoded (also notice the log locations, that will be the next issue imo) :
```python
            assert logs[0] == {"stream": "Step 1/2 : FROM alpine:latest"}
            assert logs[3] == {"stream": f'Step 2/2 : CMD echo "{random_string} 
```

alexanderankin:
there are conflicts, i would like for the tests to be clarified as well, etc.

black-snow:
Thanks, I'll take a look tonight. Will be an easy fix and I think I'll pull it out into a separate test.

What I was thinking about: Would it make sense just to pass the kwargs down to docker-py? I'd favour keywords over kwargs every time but it's basically a wrapper around docker-py, so it *might* make sense to do so. Not sure about this. On the other hand - most of the args aren't likely to change. It wouldn't be too hard to duplicate and document them.

Tranquility2:
2 notes if I may

1. What's wrong with the current test? The logic can be improved I assume, but `parametrize` makes this very slim and easy to extend. Or I can ask what the big upside on splitting into a separate test (each separate test we add is another place that will eventually needs to be maintained when updating the core functionality) 

2.  regarding "keywords over kwargs" I tend to agree but I think we should consider that it will bloat the headers and make testing and maintenance more complex. I usually like to ask ask myself does this capability has a clear/known user - in this case do you think you will use all of them? some of them? do you know of someone that will? if the answer is yes lets add the specific stuff you are going to use. Notice both `DockerClient` & `DockerContainer` emphasize on giving the minimum needed + needs discovered and anything other is left to `kwargs`, I find it very elegant as it keeps the code clean and feature oriented. we should try and avoid just duplicating if we don't have anything to add/change or the use-case is common (as good example to what's common is #615) this kind of additions can and will lead the more maintenance (and later on it will be impossible to remove or clean).

black-snow:
@Tranquility2 ofc, feedback is always welcome!

1. I think parameterized tests are fine if the parameters are for the same behaviour, if you want to cover a bunch of different inputs to test edge cases and perhaps even defaults (arguably). I'd split different behaviours (or features) into explicit tests. If a test breaks you know exactly what's wrong and the tests are living documentation.
2. True, my thoughts exactly. But should we then accept kwargs and pass them through to docker-py with a hint to check their docs? I don't _really_ need build args right now, but I'd absolutely want to be _able_ to use them when I need them.

Tranquility2:
Hi @black-snow just wanted to update I found a bug in the `buildargs` implementation https://github.com/testcontainers/testcontainers-python/issues/706 
Now fixed hopefully https://github.com/testcontainers/testcontainers-python/pull/708
you can use 
```python
with DockerImage(path=dir, tag="test", buildargs={"MY_ARG": "some_arg"}) as image:
```
if you still want to add this as a more direct flag I'm in favor, notice it now has a sperate test as you originally suggested 😄 
(If you need any help just tell me)

black-snow:
Thanks for the heads-up and the fix @Tranquility2! I'll abandon this PR. When kwargs get passed down correctly + I get a hint in the docstring (of the constructor) I'm totally happy with not having a separate kwarg :)

Tranquility2:
My plan is to add 
```python
with DockerImage().with_build_arg(key, val) as image:
```

Something like we have in the Container :)
```python
def with_build_arg(self, key: str, value: str) -> Self:
    self._build_args[key] = value
    return self
```

Hinting is also a track I'm working on and will definitely be added as well as part of https://github.com/testcontainers/testcontainers-python/pull/691 https://github.com/testcontainers/testcontainers-python/pull/692 https://github.com/testcontainers/testcontainers-python/pull/700 #702

black-snow:
I'm fine with a builder-style construct :)

Type hinting? Excellent. I was rather relating to having "kwargs will get passed throught to the underlying docker-py" right in the docstring of __init__ (or wherever it makes most sense). I remember that I was looking for a way to pass args but it just wasn't apparent how I could do that. I had to go into the code to see how. I'd rather either have an explicit kwarg or I hit ctrl+space, scan through the docstring and see that I can just pass through whatever :)

Tranquility2:
Sounds good, I'll update.

PR #616, opened 2024-06-24
chore: Update Weaviate default version to v1.25.5

antas-marcin:
This is just a small PR that updates Weaviate default version to `v1.25.5`


antas-marcin:
I need to re-test it and then will open the PR once again

PR #618, opened 2024-06-26
feat(core): Add support for ollama module

### LINKED ISSUE #617: New Container: OllamaContainer
Add support for the OllamaContainer to simplify running and testing LLMs through Ollama.

I would like to request support for a new container: OllamaContainer.

- Official Docker Hub image: [ollama/ollama](https://hub.docker.com/r/ollama/ollama)
- Ollama Blog: [Ollama Docker Image](https://ollama.com/blog/ollama-is-now-available-as-an-official-docker-image)

The generic `DockerContainer("ollama/ollama:latest")` approach is not sufficient due to several reasons:

1. **Complicated setup/configuration**: Ollama can run with GPU acceleration inside Docker containers for Nvidia GPUs. It's important to be able to check the availability of GPUs and run the container with some if possible.

2. **Model management**: There is also the need to pull a model and commit container changes into an image after pulling a model. So that the image containing the model can be reused later.

bricefotzo:
- Added a new class OllamaContainer with few methods to handle the Ollama container.

- The `_check_and_add_gpu_capabilities` method checks if the host has GPUs and adds the necessary capabilities to the container.

- The `commit_to_image` allows to save somehow the state of a container into an image so that we can reuse it, especially for the ones having some models pulled.
- Added tests to check the functionality of the new class.

> Note: I inspired myself from the java implementation of the Ollama module.

Fixes #617

### DIFF
--- a/modules/ollama/testcontainers/ollama/__init__.py
+++ b/modules/ollama/testcontainers/ollama/__init__.py
@@ -1,0 +1,79 @@
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+
+import docker
+from docker.types.containers import DeviceRequest
+
+from testcontainers.core.generic import ServerContainer
+from testcontainers.core.waiting_utils import wait_container_is_ready
+
+
+class OllamaContainer(ServerContainer):
+    def __init__(self, image="ollama/ollama:latest"):
+        self.port = 11434
+        super().__init__(port=self.port, image=image)
+        self.with_exposed_ports(self.port)
+        self._check_and_add_gpu_capabilities()
+
+    def _check_and_add_gpu_capabilities(self):
+        info = docker.APIClient().info()
+        if "nvidia" in info["Runtimes"]:
+            self.with_kwargs(device_requests=DeviceRequest(count=-1, capabilities=[["gpu"]]))
+
+    def start(self) -> "OllamaContainer":
+        """
+        Start the Ollama server
+        """
+        super().start()
+        wait_container_is_ready(self, "Ollama started successfully")
+        self._connect()
+
+        return self
+
+    def get_endpoint(self):
+        """
+        Return the endpoint of the Ollama server
+        """
+        return self._create_connection_url()
+
+    @property
+    def id(self) -> str:
+        """
+        Return the container object
+        """
+        return self._container.id
+
+    def pull_model(self, model_name: str) -> None:
+        """
+        Pull a model from the Ollama server
+
+        Args:
+            model_name (str): Name of the model
+        """
+        self.exec(f"ollama pull {model_name}")
+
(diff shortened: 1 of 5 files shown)


bricefotzo:
resolving the issue #617

modules/ollama/testcontainers/ollama/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+        """
+        Commit the current container to a new image
+
+        Args:
+            image_name (str): Name of the new image
+        """
+        docker_client = self.get_docker_client()
+        container = self.get_container()
+        existing_images = docker_client.client.images.list(name=image_name)
+        if not existing_images:
```
ilopezluna, line 75: If you don't add [this label](https://github.com/testcontainers/testcontainers-java/blob/main/modules/ollama/src/main/java/org/testcontainers/ollama/OllamaContainer.java#L70) Ryuk is going to remove the image. 
I missed that one when adding Ollama module for tc-node

bfotzo: This label namespace is protected in the python SDK. I have this issue trying to use this label:
`ValueError: The org.testcontainers namespace is reserved for internal use`

eddumelendez: Hi @bfotzo, it should be something like `docker_client.client.containers.commit` (or something like this) instead of `container.commit`

bfotzo: Hi @eddumelendez, that's what worked: `docker_client.client.containers.get(self.id).commit(repository=image_name)`following your recommendation

bricefotzo: Hi @eddumelendez, that's what worked following you recommendation:
```python
docker_client.client.containers.get(self.id).commit(repository=image_name)
```
`docker_client.client.containers` does not have the method commit directly

bricefotzo: > If you don't add [this label](https://github.com/testcontainers/testcontainers-java/blob/main/modules/ollama/src/main/java/org/testcontainers/ollama/OllamaContainer.java#L70) Ryuk is going to remove the image. I missed that one when adding Ollama module for tc-node

I fixed it, but had to change the fact that internal labels was not supposed to be changed.

eddumelendez: @bricefotzo the label shouldn't be set for the container but only when performing the `commit` operation. See https://docs.docker.com/engine/api/v1.45/#tag/Image/operation/ImageCommit

We have done something similar in Testcontainers for NodeJS, see https://github.com/testcontainers/testcontainers-node/pull/788/files#diff-212b3c0d5e84a8eb1a96cf9d0614e6d92b00feba22e9223ad86e4a66a7798d41R54

where:
- repo is a query param
- Labels is the request body

IDK if this is similar in python but we can try :D

bfotzo: @eddumelendez, that's what I also wanted to do first but didn't find in the python sdk of docker Labels assignement. But I thing laybe using the parameter conf or changes, we can achieve that. 

Let me check and try

bricefotzo: @eddumelendez , that's what I also wanted to do first but didn't find in the python sdk of docker Labels assignement. But I thing laybe using the parameter conf or changes, we can achieve that.

Let me check and try

bricefotzo: Got it!
That's the way to do in python 
```python
docker_client.client.containers.get(self.id).commit(repository=image_name, conf={"Labels": {"org.testcontainers.session-id": ""}})
```
Thanks a lot @eddumelendez !

PR #625, opened 2024-06-29
fix(modules): Mailpit Container

### LINKED ISSUE #626: New Container: Mailpit

axllent/mailpit is a full-featured SMTP mock server that's ideal for integration testing. 

- dockerhub: [axellent/mailpit](https://hub.docker.com/r/axllent/mailpit) 
- github: [axellent/mailpit](https://github.com/axllent/mailpit)
- [official docs](https://mailpit.axllent.org/docs/)

It allows for simple configs that accept any auth, and can be configured to (1) support STARTTLS and SSL; (2) accept only specific user/password combinations for auth. This covers most ways you would connect to a real SMTP server and is particularly useful for avoiding having to create a specific insecure SMTP handler for testing only—moving the mocking out of application code.

The SSL support config is non-trivial, as it requires generating a private key and a self-signed certificate, and mounting them to the container. Configuring specific user/pass auth also requires setting multiple env vars in the container in a single string with a specific format. 

For an end-user, having an SMTP server that can just be spun up that supports TLS and specific user/pass auth without understanding the internals seems ideal. If nothing else, it means you don't have to repeat the same config over and over again whenever you want to use a mailpit testcontainer.

**Other references:**

N/A

oliverlambson:
# New Container

Fixes #626 

# PR Checklist

- [x] Your PR title follows the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) syntax
  as we make use of this for detecting Semantic Versioning changes.
- [x] Your PR allows maintainers to edit your branch, this will speed up resolving minor issues!
- [x] The new container is implemented under `modules/*`
  - Your module follows [PEP 420](https://peps.python.org/pep-0420/) with implicit namespace packages
    (if unsure, look at other existing community modules)
  - Your package namespacing follows `testcontainers.<modulename>.*`
    and you DO NOT have an `__init__.py` above your module's level.
  - Your module has it's own tests under `modules/*/tests`
  - Your module has a `README.rst` and hooks in the `.. auto-class` and `.. title` of your container
  - Implement the new feature (typically in `__init__.py`) and corresponding tests.
- [x] Your module is added in `pyproject.toml`
  - it is declared under `tool.poetry.packages` - see other community modules
  - it is declared under `tool.poetry.extras` with the same name as your module name,
    we still prefer adding _NO EXTRA DEPENDENCIES_, meaning `mymodule = []` is the preferred addition
    (see the notes at the bottom)
- [x] (seems to not be needed anymore) The `INDEX.rst` at the project root includes your module under the `.. toctree` directive
- [x] Your branch is up to date (or we'll use GH's "update branch" function through the UI)

# Preferred implementation

- The current consensus among maintainers is to try to avoid enforcing the client library
  for the given tools you are triyng to implement.
- This means we want you to avoid adding specific libraries as dependencies to `testcontainers`.
- Therefore, you should implement the configuration and the waiting with as little extra as possible
- You may still find it useful to add your preferred client library as a dev dependency

### DIFF
--- a/modules/mailpit/testcontainers/mailpit/__init__.py
+++ b/modules/mailpit/testcontainers/mailpit/__init__.py
@@ -15,7 +15,7 @@ from __future__ import annotations
 import os
 import tempfile
 from datetime import datetime, timedelta, timezone
-from typing import TYPE_CHECKING, NamedTuple
+from typing import TYPE_CHECKING, Any, NamedTuple
 
 from cryptography import x509
 from cryptography.hazmat.primitives import hashes, serialization
@@ -29,9 +29,7 @@ from testcontainers.core.container import DockerContainer
 from testcontainers.core.waiting_utils import wait_for_logs
 
 if TYPE_CHECKING:
-    from typing_extensions import ParamSpec, Self
-
-    P = ParamSpec("P")
+    from typing_extensions import Self
 
 
 class MailpitUser(NamedTuple):
@@ -73,7 +71,7 @@ class MailpitContainer(DockerContainer):
             ...     # use server.sendmail(...) to send emails
     """
 
-    def __init__(  # type: ignore[no-untyped-def]
+    def __init__(
         self,
         image: str = "axllent/mailpit",
         *,
@@ -81,7 +79,7 @@ class MailpitContainer(DockerContainer):
         ui_port: int = 8025,
         users: list[MailpitUser] | None = None,
         require_tls: bool = False,
-        **kwargs,
+        **kwargs: Any,
     ) -> None:
         super().__init__(image=image, **kwargs)
         self.smtp_port = smtp_port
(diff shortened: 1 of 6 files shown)


modules/mailpit/testcontainers/mailpit/__init__.py
```diff
@@ -0,0 +1,243 @@
...
+if TYPE_CHECKING:
+    from typing_extensions import Self
+
+
+class MailpitUser(NamedTuple):
+    """Mailpit user for authentication
+
+    Helper class to define a user for Mailpit authentication.
+
+    This is just a named tuple for username and password.
```
oliverlambson, line 40: I imagine this could be a controversial addition, since it's the first class being exported from community modules that isn't a container. 

The reason I've done this is because it's a more descriptive and intuitive datatype (to me):

```diff
- users:  list[tuple[str, str]] | None
+ users:  list[MailpitUser] | None
```

modules/mailpit/testcontainers/mailpit/py.typed
oliverlambson, line 1: I've also added a `py.typed` as this module is fully type-hinted. Running mypy on it reveals the type hint issues in core, but raises no issues within this module

PR #630, opened 2024-06-30
fix(network): Now able to use Network without context, and has labels to be automatically cleaned up (#627)

### LINKED ISSUE #627: Feature: Non-context-manager API for Network
**What are you trying to do?**

I was writing some pytest fixtures to create containers, and I wanted to create a `Network` in a fixture. However, I realized that a non-context-manager interface wasn't provided for the `Network` class. I've already made the change here: https://github.com/dogukancagatay/testcontainers-python/pull/1

**Why should it be done this way?**

There is a non-context-manager API for other classes, such as `DockerContainer` and `DockerImage`, and I think that `Network` should have one for completeness, flexibility, and ease of use in some instances.

**Other references:**

Include any other relevant reading material about the enhancement.

dogukancagatay:
This PR adds `create` method to the `Network` class to enable a non-context-manager usage.

Fixes #627

### DIFF
--- a/core/testcontainers/core/docker_client.py
+++ b/core/testcontainers/core/docker_client.py
@@ -211,6 +211,10 @@ class DockerClient:
         login_info = self.client.login(**auth_config._asdict())
         LOGGER.debug(f"logged in using {login_info}")
 
+    def client_networks_create(self, name: str, param: dict):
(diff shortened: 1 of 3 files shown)


alexanderankin:
I've converted this to a draft for now as I am not sure this is the best idea

Issue #109, opened 2020-07-16
Allow option to keep the containers alive


Can-Sahin:
It would be nice to keep the docker containers alive for speeding up the test runs. Now, in every test run containers are re-created and can't keep them alive since [this block](https://github.com/testcontainers/testcontainers-python/blob/71f2a879185f772ca3f36f61ec619d1f03cccb17/testcontainers/core/container.py#L69)

```python
  def __del__(self):
        """
        Try to remove the container in all circumstances
        """
        if self._container is not None:
            try:
                self.stop()
            except:  # noqa: E722
                pass
```

removes containers when the instance variable is deleted (when the program terminates) and I cannot override it because this also runs without using `with as` block. 

I don't want my Mysql container to be recreated everytime. I am running tests very frequently and I am waiting couple seconds everytime. Pretty annoying. 

I can make a small PR if that makes sense?

HarryDunham:
Containers can be manually kept alive by simply not using a context manager and calling

stonecharioteer:
You *could* subclass it and overwrite the `__del__`, since that gives you an option to inject the behaviour that you require. That's the approach we took when we needed an alternative behaviour for the Selenium container wrapper.

tillahoffmann:
@Can-Sahin, makes sense to want to keep the container alive to speed up tests. Having said that, I'm not sure how you'd be able to connect to the existing container after the reference to the instance variable has been discarded.

If you're using pytest, you can use global fixtures to reuse the same container across all test runs within the same process. Reusing an existing container across different test runs/processes would require persisting information about the container across runs---container management should probably be outside the scope of this project.

tillahoffmann:
Closing this one for now but feel free to reopen if using fixtures with a different scope cannot address the problem.

mausch:
I figured out the same workaround mentioned in https://github.com/testcontainers/testcontainers-python/issues/109#issuecomment-748758167 , then found this issue about keeping containers alive.

I think this is worth documenting at least. I was surprised to see that my containers were getting deleted even when I explicitly didn't start them with a context. I had to dig around the code of testcontainers to figure out that it was `__del__` doing this.

[...]

PR #636, opened 2024-07-03
feat: reusable containers

matthiasschaub:
adresses #109

Todo:
- [x] Write documentation about how to reuse containers.
- [x] Warn the user if `with_reuse` in use but ryuk is disabled.
- [x] Address https://github.com/testcontainers/testcontainers-python/pull/636/changes#r1698932279

Open questions:
- [ ] Should `reuse_enable` also be configurable via environment variable?
- [ ] Currently to make reuse work ryuk needs to be disabled. Should the user do this manually? -> if so warn the user if `with_reuse` in use but ryuk is disabled.
- [ ] Current tests are sensitive to users `~/.testcontainers.properties`. This file should not be present during test run. 
- [ ] Should each container class have the possibility to state if this container can be reused? (`DockerContainer.reusable: bool = True`)


index.rst
```diff
@@ -85,6 +85,29 @@ When trying to launch a testcontainer from within a Docker container, e.g., in c
...
+
+    >>> from testcontainers.core.container import DockerContainer
+
+    >>> with DockerContainer("hello-world").with_reuse() as container:
+    ...     first_id = container._container.id
+    >>> with DockerContainer("hello-world").with_reuse() as container:
+    ...     second_id == container._container.id
+    >>> print(first_id == second_id)
+    True
+
```
matthiasschaub, line 110: Should the user be warned that by using this feature, containers need to be removed manually? (That this feature should not be used in a CI)

Also, do we need to make clear how this feature works (explaining the hash in use). -> If a container's run configuration changes, the hash changes and a new container will be used.

core/testcontainers/core/config.py
```diff
@@ -73,8 +76,19 @@ def docker_auth_config(self, value: str):
...
         return self.tc_properties.get("tc.host")
 
+    @property
+    def tc_properties_tc_host(self) -> Union[str, None]:
+        return self.tc_properties.get("tc.host")
+
+    @property
+    def tc_properties_testcontainers_reuse_enable(self) -> bool:
+        enabled = self.tc_properties.get("testcontainers.reuse.enable")
+        return enabled == "true"
```
matthiasschaub, line 90: `testcontainers.reuse.enable` (like the Java implementation) or `tc.reuse.enable` (like `tc.host`)?

`enable` (like the Java implementation) or `enabled` (like RYUK_DISABLED)?

alexanderankin:
will review and merge if you have no other plans for changes

matthiasschaub:
@alexanderankin we would welcome a review from you, thanks. They are no further plans right now except addressing what ever comes up in the code review.

core/testcontainers/core/container.py
```diff
@@ -86,8 +93,49 @@ def start(self) -> Self:
...
+            if container:
+                if container.status != "running":
+                    container.start()
+                    logger.info("Existing container started: %s", container.id)
+                logger.info("Container is already running: %s", container.id)
+                self._container = container
+            else:
+                self._start(hash_)
+        else:
+            self._start(hash_)
```
alexanderankin, line 131: can we refactor this so that is is obvious where the if clause is that triggers this?

want to make sure
1. we are doing the hash inside the clause
2. want to make it more readable - not in general but specifically for ensuring correcteness of logic that disables or enables reuse

core/testcontainers/core/container.py
```diff
@@ -96,16 +144,17 @@ def start(self) -> Self:
...
-            self._network.connect(self._container.id, self._network_aliases)
-        return self
 
     def stop(self, force=True, delete_volume=True) -> None:
         if self._container:
-            self._container.remove(force=force, v=delete_volume)
+            if self._reuse and c.tc_properties_testcontainers_reuse_enable:
+                self._container.stop()
+            else:
+                self._container.remove(force=force, v=delete_volume)
```
alexanderankin, line 157: hm, isnt the point to not even stop it so it is warm for next run? i guess if people are using the explicit api then whatever. I do see a bit of a mirror with start so i guess it will just have to be consistent and maybe clear in docs.

index.rst
```diff
@@ -118,6 +117,32 @@ Fetching passwords from cloud providers:
     GCP_PASSWORD = $(gcloud auth print-access-token)
     AZURE_PASSWORD = $(az acr login --name <registry-name> --expose-token --output tsv)
 
+Reusable Containers (Experimental)
+----------------------------------
+
+Containers can be reused across consecutive test runs. To reuse a container, the container configuration must be the same.
+
+Containers that are set up for reuse will not be automatically removed. Thus, those containers need to be removed manually.
```
alexanderankin, line 125: "...removed manually."

maybe add:
"In re-usable mode, the 'stop' api on a container will now 'stop' a container, rather than 'remove' it"

index.rst
alexanderankin: seems like you have added these comments to the doc, i think that is fine. the hash would be great to add as users would benefit from knowing exactly what is hashed.

* `self.image,`
* `self._command,`
* `self.env,`
* `self.ports,`
* `self._name,`
* `self.volumes,`
* `str(tuple(sorted(self._kwargs.items()))),` - this may fail and why i want to have this be tucked away inside an obviously readable `if` block

core/testcontainers/core/container.py
```diff
@@ -86,8 +93,49 @@ def start(self) -> Self:
...
+                logger.info("Container is already running: %s", container.id)
+                self._container = container
+            else:
+                self._start(hash_)
+        else:
+            self._start(hash_)
+
+        if self._network:
+            self._network.connect(self._container.id, self._network_aliases)
+        return self
```
alexanderankin, line 135: apparently this part is also fairly jank and we should remove/rework so as a note to myself i can only do that after this pr merges

core/testcontainers/core/config.py
eddumelendez: Let's keep the same key as Java implementation `testcontainers.reuse.enable`.

kiview: Yes, this is important. This key is used across most language implementations and the way tc-java does is considered the reference implemetation.

core/testcontainers/core/container.py
kiview: In other languages, having a reusable container *does not* change the contract of the `stop()` method. This is obviously something that needs to be considered to make this a full fledged use case, but as of now, I would suggest we start with an experimental reusable implementation, that mirrors the Java implementation.

kiview:
Thanks for looking into this @matthiasschaub 👋 
We have no clean spec for this features in Java / across languages, so for now I would suggest, we mostly mirror the Java implementation, including its limitations. We can sync on a cross language spec in the future that provides a better DX, but for now I would strongly favor an implementation that mirrors Java.

I already left some comments within the PR.

> Currently to make reuse work ryuk needs to be disabled. Should the user do this manually?

Given the above, if a container has `reusable` set, it *must not* be registered with the Ryuk cleanup label (see [tc-java](https://github.com/testcontainers/testcontainers-java/blob/c491f6aa543a4657e53578e8a0581ab375f8c105/core/src/main/java/org/testcontainers/containers/GenericContainer.java#L432-L435)).

core/testcontainers/core/container.py
```diff
@@ -86,8 +93,49 @@ def start(self) -> Self:
...
+        args = (
+            self.image,
+            self._command,
+            self.env,
+            self.ports,
+            self._name,
+            self.volumes,
+            str(tuple(sorted(self._kwargs.items()))),
+        )
+        hash_ = hashlib.sha256(bytes(str(args), encoding="utf-8")).hexdigest()
```
kiview, line 108: Ideally we use the *full* container create request as the hash input. In tc-java, this is the `CreateContainerCmd` from docker-java, I guess we have some equivalent request object from the Docker Python SDK somewhere available?

core/testcontainers/core/config.py
matthiasschaub: Okay, then let's keep it as implemented.

core/testcontainers/core/container.py
matthiasschaub: Okay, I see. I updated the code and documentation (how to use reusable containers) to not change the contract of the `stop()` method.

index.rst
matthiasschaub: After [this](https://github.com/testcontainers/testcontainers-python/pull/636#discussion_r1701699697) discussion, the stop method has not been changed.

matthiasschaub:
> Thanks for looking into this @matthiasschaub 👋 We have no clean spec for this features in Java / across languages, so for now I would suggest, we mostly mirror the Java implementation, including its limitations. We can sync on a cross language spec in the future that provides a better DX, but for now I would strongly favor an implementation that mirrors Java.
> 
> I already left some comments within the PR.
> 
> > Currently to make reuse work ryuk needs to be disabled. Should the user do this manually?
> 
> Given the above, if a container has `reusable` set, it _must not_ be registered with the Ryuk cleanup label (see [tc-java](https://github.com/testcontainers/testcontainers-java/blob/c491f6aa543a4657e53578e8a0581ab375f8c105/core/src/main/java/org/testcontainers/containers/GenericContainer.java#L432-L435)).

Thanks for the review! I agree. It is sensible to follow the Java implementation.

In commit 1ea9ed16a0dd73e11d762808aed5655d44e270be I do not create a Reaper instance during container start-up if reuse is enabled and container has been started with `with_reuse`. This works as expected as long as no other container is started without `with_reuse`. In that case a Reaper instances is created which will remove the reuse container as well.
Is there a way to explicitly exclude a container from the Reapers routine?

core/testcontainers/core/container.py
matthiasschaub: Thanks for the feedback! I will revisit this part next week and try to improve upon it.

core/testcontainers/core/container.py
matthiasschaub: Yeah, that would be the ideal solution. Unfortunately, I could not find an equivalent function to `CreateContainerCmd` in the Docker SDK for Python: Not by going through the documentation and not by browsing the code base.

core/testcontainers/core/container.py
matthiasschaub: @alexanderankin I moved the generation of the hash inside the if-clause and removed passing the `hash_` to start if *reuse* is not in use. I think that makes it better readable in general.

matthiasschaub:
Please let me know if I can do anything to advance this PR towards merging :)

index.rst
matthiasschaub: Maybe use something like https://github.com/knowsuchagency/picocache/blob/main/picocache/utils.py#L9 for making the hash key

core/testcontainers/core/config.py
```diff
@@ -125,9 +128,19 @@ def docker_auth_config(self, value: str) -> None:
...
     @property
+    def tc_properties_tc_host(self) -> Union[str, None]:
+        return self.tc_properties.get("tc.host")
+
+    @property
+    def tc_properties_testcontainers_reuse_enable(self) -> bool:
+        enabled = self.tc_properties.get("testcontainers.reuse.enable")
+        return enabled == "true"
+
     def timeout(self) -> int:
```
gpkc, line 144: I just noticed that `timeout` here had the `@property` decorator before. Was it removed by accident?

matthiasschaub: Yes, this was an accident.

chbndrhnns:
Is this PR still intended to be merged?

core/testcontainers/core/container.py
matthiasschaub: Instead of `sh256` its probably cheaper and faster to use `zlib.crc32`.

index.rst
matthiasschaub: Before running `str(tuple(sorted(self._kwargs.items())))`, `self._kwargs` is now beeing checked for `None`

matthiasschaub:
When reviewing, please also check open questions stated in the PR description. Thanks :)

iirekm:
why it takes so long? I would love to have reuse just like in Java

dantetemplar:
Upvote

jacky-micro1:
upvote

matthiasschaub:
I incorporated latest changes from `main` into this branch. I would be curious if the CI runs through successfully.

PR #647, opened 2024-07-12
fix(core): Improve private registry support (tolerate not implemented fields in DOCKER_AUTH_CONFIG)

Tranquility2:
Continuing #562, got some feedback regarding an issue with unsupported use cases.
In this PR we will try to:
1. Map the use cases
2. Raise a warning regarding unsupported uses cases (hopefully they will be added later)
3. Address/Fix the issue where unsupported JSON schema for `DOCKER_AUTH_CONFIG` leads to an error

As always any feedback will be much appreciated.
_Please note this PR does not implement all use-cases just does a better job at preparing and handling them for now_

### DIFF
--- a/core/testcontainers/core/auth.py
+++ b/core/testcontainers/core/auth.py
@@ -6,7 +6,7 @@ from typing import Optional
 
 DockerAuthInfo = namedtuple("DockerAuthInfo", ["registry", "username", "password"])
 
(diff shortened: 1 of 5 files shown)


alexanderankin:
since this does not actually implement the features i am going to merge this as a fix

PR #669, opened 2024-08-09
feat(compose): add ability to get docker compose config

g0di:
This PR add a new function to the `testcontainers.compose.DockerComposer` class, `get_config` which use `docker compose config` command for resolving and returning the actual docker compose configuration.

This can be useful for example if you want to retrieve a connection string you pass to your app in your docker compose in order to connect to your database service instead of copy pasting it from your compose file into your tests.

Also note thats its way easier to rely on docker compose config to get you the config than trying to manually find, read and merge compose files in specified context (I tried it first ...).

About the tests I mostly ensured the docker compose command was as expected. This is because the config produced by the docker compose can not always reflect exactly what is in the file. There is some normalization/resolving which is done (even when you pass all flags to disable them). But anyway, I'm not sure its a good idea to actually test the behavior of the docker config command itself. 

Let me know what you think of it!

### DIFF
--- a/core/testcontainers/compose/compose.py
+++ b/core/testcontainers/compose/compose.py
@@ -1,3 +1,4 @@
+import json
 from dataclasses import asdict, dataclass, field, fields
 from functools import cached_property
 from json import loads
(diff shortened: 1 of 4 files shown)


alexanderankin:
im not sure i like the change to the public api so i marked as experimental and lets see where it goes.

g0di:
Fair enough, thanks for your review ;)

PR #671, opened 2024-08-09
docs: Documentation fix for ServerContainer

sasa-buklijas:
Used same wording as for modules/aws/testcontainers/aws/aws_lambda.py


Tranquility2 (approved):

alexanderankin:
are we going to fix generic module here or later?

Issue #672, opened 2024-08-11
New Container: Db2


icarrusss:
**What is the new container you'd like to have?**

IBM Db2 database.
Container: [ibmcom/db2](https://hub.docker.com/r/ibmcom/db2)
Documentation: [deployments-db2-community-edition-docker](https://www.ibm.com/docs/en/db2/11.5?topic=deployments-db2-community-edition-docker)

**Why not just use a generic container for this?**

I noticed there is a PR https://github.com/testcontainers/testcontainers-python/pull/321 for the Db2 support, but it has been closed.

IBM Db2 is a widely used relational database, and it should be supported as a dedicated container.
The Db2 container requires specific setup configurations as listed in https://hub.docker.com/r/ibmcom/db2, and the wait strategry involves checking the availability of the database or some specific keywords in the container log.

**Other references:**

Include any other relevant reading material about the enhancement.

PR #673, opened 2024-08-11
fix: Add Db2 support

icarrusss:
Add Db2 support https://github.com/testcontainers/testcontainers-python/issues/672.


alexanderankin:
30 minutes to run tests: I feel like its creating a monster. however, it would probably be worse not be able to use a testcontainer. for any future reference, there is absolutely no need to:

1. edit other modules when submitting a new one
2. duplicate the password test.

Issue #665, opened 2024-08-05
Feature: add DockerContainer.with_copy_file_ton_container


Lenormju:
**What are you trying to do?**

Describe the intention of the enhancement.

I'd like to add a file to the container I'm starting, without using volumes.

**Why should it be done this way?**

Feature-parity with the Java version : https://java.testcontainers.org/features/files/#copying-files

**Other references:**

It you deem it interesting to implement, as I do, I can implement it.

Lenormju:
I have found this remark in a PR : https://github.com/testcontainers/testcontainers-python/pull/392#discussion_r1390988566
Indeed, this PR was not the right place to do it, but to implement it in a new one now would be perfect 😄

anuraaga:
In case anyone lands on this and wishes for copying (I use it to pass GCP creds to cloud sql proxy container, volume mount doesn't work on Linux), I use this rough workaround, the key point is to set restart policy on failure to allow it to fail as expected on missing creds file until the copy completes.

```
# Roughly based on proposed logic in https://github.com/testcontainers/testcontainers-python/pull/676/files
def copy_file_to_container(container: DockerContainer, host_path: str, container_path: str, mode: int):
    data = BytesIO()

    def set_mode(tarinfo: tarfile.TarInfo):
        tarinfo.mode = mode
        return tarinfo

    with tarfile.open(fileobj=data, mode='w') as tar:
        tar.add(host_path, arcname=container_path, filter=set_mode)

    data.seek(0)
    if not container._container.put_archive('/', data):
        raise Exception('Failed to copy file to container')

with (
        DockerContainer('gcr.io/cloud-sql-connectors/cloud-sql-proxy:2.14.3')
        .with_command(
            '--address 0.0.0.0 --health-check --http-address 0.0.0.0 foo'
        )
        .with_exposed_ports(5432, 9090)
        .with_env(environment_vars.CREDENTIALS, '/creds/creds.json')
        # Currently there is no way to copy a file before container starts, so we workaround by allowing the container
        # to restart on failure, which will stop when the copy has completed.
        # https://github.com/testcontainers/testcontainers-python/pull/676#discussion_r1934888327
        .with_kwargs(restart_policy={'Name': 'on-failure', 'MaximumRetryCount': 100})
    ) as container:
        copy_file_to_container(container, creds_path, '/creds/creds.json', 0o644)
        wait_for_logs(container, 'The proxy has started successfully and is ready for new connections!')
```

[...]

PR #676, opened 2024-08-12
feat: add copying files to/from container

mgorsk1:
solves https://github.com/testcontainers/testcontainers-python/issues/665

### DIFF
--- a/core/testcontainers/core/container.py
+++ b/core/testcontainers/core/container.py
@@ -1,7 +1,10 @@
 import contextlib
+import io
+import tarfile
+from pathlib import Path
 from platform import system
 from socket import socket
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING, Optional, Tuple
 
 import docker.errors
 from typing_extensions import Self
@@ -50,6 +53,7 @@ class DockerContainer:
         self._network: Optional[Network] = None
         self._network_aliases: Optional[list[str]] = None
         self._kwargs = kwargs
+        self._files: list[Tuple[Path, Path]] = []
 
     def with_env(self, key: str, value: str) -> Self:
         self.env[key] = value
@@ -76,6 +80,33 @@ class DockerContainer:
         self._kwargs = kwargs
         return self
 
+    def with_copy_file_to_container(self, source_file: Path, destination_file: Path) -> Self:
+        self._files.append((source_file, destination_file))
+
+        return self
+
+    def copy_file_from_container(self, container_file: str, destination_file: str) -> str:
+        tar_stream, _ = self._container.get_archive(container_file)
+
+        for chunk in tar_stream:
+            with tarfile.open(fileobj=io.BytesIO(chunk)) as tar:
+                for member in tar.getmembers():
+                    with open(destination_file, 'wb') as f:
+                        f.write(tar.extractfile(member).read())
+
+        return destination_file
+
+    @staticmethod
+    def _put_file_in_container(container, source_file: Path, destination_file: str):
+        data = io.BytesIO()
+
+        with tarfile.open(fileobj=data, mode='w') as tar:
+            tar.add(source_file, arcname=destination_file)
+
+        data.seek(0)
+
+        container.put_archive("/", data)
+
     def maybe_emulate_amd64(self) -> Self:
         if is_arm():
             return self.with_kwargs(platform="linux/amd64")
@@ -88,7 +119,7 @@ class DockerContainer:
         logger.info("Pulling image %s", self.image)
         docker_client = self.get_docker_client()
         self._configure()
-        self._container = docker_client.run(
+        self._container: Container = docker_client.run(
             self.image,
             command=self._command,
             detach=True,
@@ -101,6 +132,12 @@ class DockerContainer:
         logger.info("Container started: %s", self._container.short_id)
         if self._network:
             self._network.connect(self._container.id, self._network_aliases)
+
+        for file in self._files:
+            source, destination = file[0], file[1]
+
+            DockerContainer._put_file_in_container(self._container, source, destination)
+
         return self
 
     def stop(self, force=True, delete_volume=True) -> None:


alexanderankin:
i usually prefer using `typing.Union[str, os.PathLike]` - if you wanna use strings with this as well as paths

alexanderankin:
one other thing to consider is that there is this great interface "Transferable" in other language implementations-- does this PR make it harder or easier to add this to this implementation as well.

the interface (found [here](https://github.com/testcontainers/testcontainers-java/blob/4b3d08ecff0923e37caac83e58807076c0aa9d44/core/src/main/java/org/testcontainers/images/builder/Transferable.java#L68-L86)) is basically:

```python
class Transferable(abc.ABC):
  @abstractmethod
  def transfer_to(content_stream, destination: typing.Union[str, os.PathLike]):
    ...
```

where content_stream is some bytesio or something pythonic for a place where you can write bytes to.

alexanderankin:
when i tried to do it, i accidentally interpreted the interface a bit too widely and instead of passing just the content output stream, i passed the whole container as that first argument.

https://github.com/testcontainers/testcontainers-python/commit/4b7db74176c0f00ca275bc39409ebb891a93cbb9 - misc functions, next commit doesnt make sense without reading first

https://github.com/testcontainers/testcontainers-python/commit/bbc44ec8e6d0ce35d40466f1d636106d75e0b197 - the commit where i add the transferable stuff

alexanderankin:
so im tiny bit hesitant to merge without a plan for adding transferable on top of this

mgorsk1:
do we need it to be 1:1 with java implementation though? any binary object now can be passed here using tempfile.NamedTemporaryObject:

with tempfile.NamedTemporaryObject() as f:

    f.write(binary_data)
    DockerContainer("nginx").with_copy_file_to_container(f.name, "/tmp/target.data").start()

alexanderankin:
we dont need to be 1-1 but what if instead of a file you had a string? i think it is beneficial to not force people to create temporary files.

mgorsk1:
refactored for Transferable although creating temp files cannot be avoided, now instead of asking users for 3 extra lines we add some complexity to codebase here. working example

input:

```
from pathlib import Path

from testcontainers.core.container import DockerContainer, Transferable
from testcontainers.core.network import Network

c = DockerContainer('nginx')

network = Network()

c \
    .with_copy_file_to_container(Transferable(b'asd', Path('/tmp/mariusz.md1'))) \
    .with_copy_file_to_container(Transferable(Path('/Users/mariusz/Documents/Programming/testcontainers-python/README.md'),Path('/tmp/mariusz.md2'))) \
    .start()

c2 = DockerContainer('nginx')

print(c._container.exec_run('head /tmp/mariusz.md1'))
print(c._container.exec_run('head /tmp/mariusz.md2'))
```

output

```
ExecResult(exit_code=0, output=b'asd')
ExecResult(exit_code=0, output=b'[![Poetry](https://img.shields.io/endpoint?url=https://python-poetry.org/badge/v0.json)](https://python-poetry.org/)\n[![Ruff](https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/astral-sh/ruff/main/assets/badge/v2.json)](https://github.com/astral-sh/ruff)\n![PyPI - Version](https://img.shields.io/pypi/v/testcontainers)\n[![PyPI - License](https://img.shields.io/pypi/l/testcontainers.svg)](https://github.com/testcontainers/testcontainers-python/blob/main/LICENSE)\n[![PyPI - Python Version](https://img.shields.io/pypi/pyversions/testcontainers.svg)](https://pypi.python.org/pypi/testcontainers)\n[![codecov](https://codecov.io/gh/testcontainers/testcontainers-python/branch/master/graph/badge.svg)](https://codecov.io/gh/testcontainers/testcontainers-python)\n![Core Tests](https://github.com/testcontainers/testcontainers-python/actions/workflows/ci-core.yml/badge.svg)\n![Community Tests](https://github.com/testcontainers/testcontainers-python/actions/workflows/ci-community.yml/badge.svg)\n[![Docs](https://readthedocs.org/projects/testcontainers-python/badge/?version=latest)](http://testcontainers-python.readthedocs.io/en/latest/?badge=latest)\n\n')

```

Lenormju:
@mgorsk1 I have opened #677 to solve the same issue (I did not see you already had opened one 😓).
Yours have the `Transferable` feature. But I have the `read_only` feature.
And they are not implemented the same way :

* I am passing the paths in the `HostConfig.Bind` to the docker_client.create (so the file exist before the container starts)
* You are writing the files using the `docker.models.containers.Container.put_archive` method (which works after the container has been started)

Are our two approaches complementary ?
Reading again [the Java documentation for copying files](https://java.testcontainers.org/features/files/#copying-files), there does not seem to be a difference, at the API level at least, between copying to a container **before** or **after** its startup.

mgorsk1:
I think copying before can already be solved by with_volume 😅

alexanderankin:
committed some code here - https://github.com/testcontainers/testcontainers-python/compare/main...feat/665-with-copy-file-to-container

alexanderankin:
rebased here https://github.com/testcontainers/testcontainers-python/compare/feat/665-with-copy-file-to-container_rebased

alexanderankin:
since this is not a high priority feature this can probably wait for 1) more api refinement - feedback welcomed on my suggestions above 2) more of my free time, as I'm not sure ill be able to spend too much more time on this this week (just spent my entire week's time budget on this today).

core/testcontainers/core/container.py
```diff
@@ -115,6 +147,10 @@ def start(self) -> Self:
         )
 
         logger.info("Container started: %s", self._container.short_id)
+
+        for transferable in self._files:
```
anuraaga, line 151: I found this PR noticing that copying files isn't supported, a bit surprising since it's a pretty important one and found in at least many of the other languages.

I noticed this implementation doesn't seem to match others, though this is probably because of how the general docker running is implemented. Generally, the container is created, then files copied, then the container started, so the files are present when the command runs.

https://github.com/testcontainers/testcontainers-java/blob/1404c4429c0cb98fe46534bf33632d25dc5309e4/core/src/main/java/org/testcontainers/containers/GenericContainer.java#L430

Here, the files are just copied after startup so can't be used by the entrypoint - perhaps there's some use case for this but I don't know how common it is. Note that volume mounting is an alternative but doesn't always work due to e.g. permission mode issues - copying is generally more robust.

Given the behavior would be so different from other implementations of testcontainers, I think this PR can cause a lot of confusion so probably needs to rework the lifecycle like that.

core/testcontainers/core/container.py
```diff
@@ -78,6 +83,33 @@ def with_kwargs(self, **kwargs) -> Self:
...
+                        f.write(tar.extractfile(member).read())
+
+        return destination_file
+
+    @staticmethod
+    def _put_data_in_container(container, transferable: Transferable):
+        data = io.BytesIO()
+
+        with transferable as f, tarfile.open(fileobj=data, mode="w") as tar:
+            tar.add(f.input_path, arcname=f.output_path)
```
anuraaga, line 107: IIUC, the temporary file could be avoided if using `tar.addfile` - it is somewhat more tedious to set up a `TarInfo` but maybe worth it to skip the temp file

anuraaga: Also, I think all the copy_file APIs in testcontainers accept a file mode. With the current pattern, it can be something like `filter=lambda tarinfo: tarinfo.mode = foo`, but if switching to `addfile` it could be set directly to the `TarInfo`

core/testcontainers/core/container.py
rhoban13: Making this happen during `start` doesn't feel like the right interface here.  While there's definitely good reasons to copy in files just after it starts, as proposed here that is *the only* time we can copy files into the container.  

I'd propose instead creating a `copy_file_into_container`, which would basically be the inverse of `copy_file_from_container` proposed above.  It could be implemented by simply calling `get_archive`, but provide an interface which used `os.PathLike`.

Tranquility2:
Hi, just a quick checkup, is this still relevant now that we have `container.copy_into_container(my_file, "/tmp/my_file.txt")` ready?

Issue #681, opened 2024-08-14
Bug: `testcontainers.core.waiting_utils.wait_for_logs` keeps waiting despite the container has exited.


ArthurBook:
**Describe the bug**

The `testcontainers.core.waiting_utils.wait_for_logs` keeps waiting despite the container has exited. This can be confusing if there is a long wait time.

**To Reproduce**
```python
from testcontainers.core import container, waiting_utils

if __name__ == "__main__":
    waiting_utils.wait_for_logs(
        container.DockerContainer("flyway/flyway").start(),
        r"Successfully applied \d+ migrations to schema",
        timeout=large_n, # 
    )

# >> Raised Exception has occurred: TimeoutError Container did not emit logs satisfying predicate in `large_n` seconds
```

**Runtime environment**
```bash
Python:         3.11.9
testcontainers  4.7.2
Implementation: CPython
```

PR #682, opened 2024-08-14
fix: wait_for_logs can now fail early when the container stops

ArthurBook:
Addresses my suggestion made in [issue 681](https://github.com/testcontainers/testcontainers-python/issues/681).

This PR adds a flag that checks is the status is not `running` and raises a `RuntimeError` to avoid waiting for logs after the container already has exited. The idea is to save wait time when there is a long startup time in case the container fails early.

```python
from testcontainers.core import container, waiting_utils

if __name__ == "__main__":
    waiting_utils.wait_for_logs(
        container.DockerContainer("flyway/flyway").start(),
        r"Successfully applied \d+ migrations to schema",
        timeout=10,
        raise_on_exit=True,
    )

# > RuntimeError(f"Container exited before emitting logs satisfying predicate")
# ( Raised almost immediately ) 
```

### DIFF
--- a/core/testcontainers/core/waiting_utils.py
+++ b/core/testcontainers/core/waiting_utils.py
@@ -83,6 +83,7 @@ def wait_for_logs(
     timeout: float = config.timeout,
     interval: float = 1,
     predicate_streams_and: bool = False,
+    raise_on_exit: bool = False
     #
 ) -> float:
     """
@@ -117,4 +118,6 @@ def wait_for_logs(
             return duration
         if duration > timeout:
             raise TimeoutError(f"Container did not emit logs satisfying predicate in {timeout:.3f} " "seconds")
+        if raise_on_exit and container.get_wrapped_container().status != 'running':
+            raise RuntimeError(f"Container exited before emitting logs satisfying predicate")
         time.sleep(interval)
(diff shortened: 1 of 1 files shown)


alexanderankin:
why is true is not the default, any background on this?

ArthurBook:
> why is true is not the default, any background on this?

I think it would be a good default.
Kept it False just to not change existing functionality without explicit opt-in.

alexanderankin:
i would like to roll it out everywhere else, wonder what would be a good strategy for that, unless i am overthinking it, in which case we should just do it.

kiview:
TBH, I think the `false` behavior is more a bug than actual expected behavior. If I were to make the call, I would not only make `true` the default, but also don't make it user configurable at all (can any of you come up with a use case were the `false` behavior is useful?).

alexanderankin:
pr title = "fix: wait_for_logs can now fail early when the container stops"

next pr title = ${pr title/can/WILL/}

ArthurBook:
Hi! 
Agree with above, for all our use cases, the expected behavior would be to raise RuntimeError if the container exited. I think its a good idea to drop the option.

Another option is to allow user to define the set of [`Container.status`](https://docker-py.readthedocs.io/en/stable/containers.html?highlight=status#docker.models.containers.Container.status) that (don't) cause a fail. This would allow for a bit more configurability. WDYT?

PR #691, opened 2024-08-22
fix(core): Typing in auth

Tranquility2:
Lets try and continue the journey from #305 
(the goal is to get core marked as typed)

```
poetry run mypy --config-file pyproject.toml core/tests/test_auth.py core/testcontainers/core/auth.py
Success: no issues found in 2 source files
```


Tranquility2:
I know https://github.com/testcontainers/testcontainers-python/pull/504 also exists but I feel we need to make it into smaller iteration to get the a goal, treating each core component as a milestone doesn't mean it will all be perfect (time and other commits are bound to make a mess) but it should allow us to get much closer so that the the last mile will be much simpler.

PR #699, opened 2024-09-08
fix(tests): Missing artifacts (include-hidden-files)

Tranquility2:
`upload-artifact` updated behavior: Hidden Files will be excluded by default 
https://github.com/actions/upload-artifact/issues/602

Need to include 
```yaml
include-hidden-files: true
```


Tranquility2:
@alexanderankin this is needed to make the CI tests pass.
Sadly https://github.com/actions/upload-artifact/issues/602 did a breaking change without adhering to semantic versioning..

alexanderankin:
fascinating

Issue #706, opened 2024-10-01
Bug: Docker image build kwargs is not working


Tranquility2:
**Describe the bug**

When using kwargs in the Image API, the params are not passed into the `build`

**To Reproduce**

For example:
```python
with DockerImage(path="./", tag="new:test", buildargs={"PIP_EXTRA_INDEX_URL": PIP_EXTRA_INDEX_URL}) as image:
```
The `PIP_EXTRA_INDEX_URL` fails to register in the docker image build process:
```
E               docker.errors.BuildError: The command '/bin/sh -c pip3 install --no-cache-dir -r requirements.txt --extra-index-url ${PIP_EXTRA_INDEX_URL}' returned a non-zero code: 2
```

Tranquility2:
```python
class DockerImage:
...
    def __init__(
        self,
        path: Union[str, PathLike],
        docker_client_kw: Optional[dict] = None,
        tag: Optional[str] = None,
        clean_up: bool = True,
        dockerfile_path: Union[str, PathLike] = "Dockerfile",
        no_cache: bool = False,
        **kwargs,
    ) -> None:
...
    def build(self, **kwargs) -> Self:
        logger.info(f"Building image from {self.path}")
        docker_client = self.get_docker_client()
        self._image, self._logs = docker_client.build(
            path=str(self.path), tag=self.tag, dockerfile=self._dockerfile_path, nocache=self._no_cache, **kwargs
        )
        logger.info(f"Built image {self.short_id} with tag {self.tag}")
        return self
...
    def __enter__(self) -> Self:
        return self.build()
```
As can be seen `kwargs` is not passed/used correctly...
This is ofc my bad, sorry.
Will create a fix asap.

maybe also take the time to add something like https://github.com/testcontainers/testcontainers-python/pull/614

Tranquility2:
### Workaround
Just use the lower level methods
```python
def test_something():
    image = DockerImage(path="./", tag="new:test")
    image.build(buildargs={"PIP_EXTRA_INDEX_URL": PIP_EXTRA_INDEX_URL})
    with AWSLambdaContainer(image=image, port=8080) as func:
       # Do Something with the container
    image.remove()
```

black-snow:
I feel like this was my original issue with build args exactly :D

PR #708, opened 2024-10-01
fix(core): Add kwargs to image build

Tranquility2:
Fix: #706, https://github.com/testcontainers/testcontainers-python/pull/614

Now when using kwargs in the Image API, the params are passed correctly into the build

```python
with DockerImage(path=dir, tag="test", buildargs={"MY_ARG": "some_arg"}) as image:
```

Added relevant test + updated docstring to better reflect the usage


black-snow:
Can we perhaps add a hint to the docstring that kwargs will be passed down to docker-py? That would pretty much solve https://github.com/testcontainers/testcontainers-python/pull/614 for me.

Tranquility2:
@black-snow Added as suggested.

alexanderankin (approved):

Issue #537, opened 2024-04-12
Bug: RYUK container startup failure when running Docker in rootless mode


keneanung:
**Describe the bug**

When Docker is run in [rootless mode](https://docs.docker.com/engine/security/rootless/), the `ryuk` fails to start as the docker socket mounted as a volume has the wrong permissions (`nobody:nobody`). This can be fixed be mounting the correct socket at `/run/user/$(id -u)/docker.sock`.

The overwrite can be done by setting the environment variable `TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE` to the above mentioned socket.

Ideally, this would be detected and configured automatically (ie. using `Server.Security Options.rootless` and `Endpoints.docker.Host` of the active context or the `DOCKER_HOST` environment variable).

If this is not an option, please consider documenting this case more in detail.

**To Reproduce**

Configure Docker to run in rootless mode (see https://docs.docker.com/engine/security/rootless/) and run the following snippet:

```python
>>> from testcontainers.postgres import PostgresContainer
>>> with PostgresContainer() as container:
...   assert True
```

**Runtime environment**

Provide a summary of your runtime environment. Which operating system, python version, and docker version are you using? What is the version of `testcontainers-python` you are using? You can run the following commands to get the relevant information.

<details>

```bash
# Get the operating system information (on a unix os).
$ uname -a
Linux FE-C-012RG 6.5.0-1019-oem #20-Ubuntu SMP PREEMPT_DYNAMIC Mon Mar 18 17:38:55 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
# Get the python version.
$ python --version
Python 3.12.2
# Get the docker version and other docker information.
$ docker info
Client: Docker Engine - Community
 Version:    26.0.1
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.13.1
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.26.1
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 19
 Server Version: 26.0.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: false
  userxattr: true
 Logging Driver: json-file
[...]

PR #710, opened 2024-10-02
fix(core): Fix retrieval of the docker socket path when using rootless docker

vemonet:
Currently the docker socket path used by `ryuk` is not working for rootless docker. We need to manually set it through the `TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE` environment variable. 

When running locally the `DOCKER_HOST` is always the full URL to the docker socket, e.g. on linux/macos, it's the socket path prefixed with `unix://` and `npipe://` on windows. So we can easily extract the socket path using urllib parse:

```python
from urllib.parse import urlparse
docker_socket = urlparse(DOCKER_HOST).path
```

I improved the `TestcontainersConfiguration` so that the docker socket is inferred from the docker host when defined (and when the `RYUK_DOCKER_SOCKET` env is not explicitly defined)

The advantage of this approach is that it is automatically inferred from the `DOCKER_HOST` most of the time without the need for tedious manual configuration. And users can still explicitly define it for edge use-cases like before (e.g. when using docker over the network)

I moved `get_docker_host()` to make it part of the `TestcontainerConfig`, it was used at only 2 places, so it did not required many changes. And it makes sense to be part of the config

## Alternative solution

An alternatively solution has been proposed in https://github.com/testcontainers/testcontainers-python/issues/537. But it would rely on many uncertain and moving parts:
- the path to the docker socket in rootless docker to always look like this `"/run/user/{user_id}/docker.sock"` which might change in the future or be different depending on the OS and config
- being able to get the user ID somehow, on linux we can do it through the `UID` env variable, not sure if this works on macos, and it would probably not work on windows.

This solution could be implemented in `container.py`, e.g.:

```python
 def _create_instance(cls) -> "Reaper":
        logger.debug(f"Creating new Reaper for session: {SESSION_ID}")
        client = DockerClient()
        info = client.info()
        sec_opts = info.get('SecurityOptions') or tuple()
        is_rootless = any('rootless' in s for s in sec_opts)
        docker_socket = c.ryuk_docker_socket
        user_id = os.getenv('UID')
        if is_rootless and user_id:
            docker_socket = f"/run/user/{user_id}/docker.sock"
        Reaper._container = (
            DockerContainer(c.ryuk_image)
            .with_name(f"testcontainers-ryuk-{SESSION_ID}")
            .with_exposed_ports(8080)
            .with_volume_mapping(docker_socket, "/var/run/docker.sock", "rw")
            .with_kwargs(privileged=c.ryuk_privileged, auto_remove=True)
            .with_env("RYUK_RECONNECTION_TIMEOUT", c.ryuk_reconnection_timeout)
            .start()
        )
```

And would require to add `client.info()` in `docker_client.py`

```python
def info(self) -> Any:
        return self.client.info()
```

Fixes https://github.com/testcontainers/testcontainers-python/issues/537

@alexanderankin

### DIFF
--- a/core/testcontainers/core/config.py
+++ b/core/testcontainers/core/config.py
@@ -4,6 +4,7 @@ from os import environ
 from os.path import exists
 from pathlib import Path
 from typing import Optional, Union
+from urllib.parse import urlparse
 
 MAX_TRIES = int(environ.get("TC_MAX_TRIES", 120))
 SLEEP_TIME = int(environ.get("TC_POOLING_INTERVAL", 1))
@@ -12,7 +13,7 @@ TIMEOUT = MAX_TRIES * SLEEP_TIME
 RYUK_IMAGE: str = environ.get("RYUK_CONTAINER_IMAGE", "testcontainers/ryuk:0.8.1")
 RYUK_PRIVILEGED: bool = environ.get("TESTCONTAINERS_RYUK_PRIVILEGED", "false") == "true"
 RYUK_DISABLED: bool = environ.get("TESTCONTAINERS_RYUK_DISABLED", "false") == "true"
-RYUK_DOCKER_SOCKET: str = environ.get("TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE", "/var/run/docker.sock")
+RYUK_DOCKER_SOCKET: Optional[str] = environ.get("TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE", None)
 RYUK_RECONNECTION_TIMEOUT: str = environ.get("RYUK_RECONNECTION_TIMEOUT", "10s")
 TC_HOST_OVERRIDE: Optional[str] = environ.get("TC_HOST", environ.get("TESTCONTAINERS_HOST_OVERRIDE"))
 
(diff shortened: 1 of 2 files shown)


core/testcontainers/core/config.py
```diff
@@ -49,7 +50,6 @@ class TestcontainersConfiguration:
     ryuk_image: str = RYUK_IMAGE
     ryuk_privileged: bool = RYUK_PRIVILEGED
     ryuk_disabled: bool = RYUK_DISABLED
-    ryuk_docker_socket: str = RYUK_DOCKER_SOCKET
```
alexanderankin, line 52: we need to leave this configurable because otherwise people want to monkeypatch our modules

alexanderankin:
> "And would require to add client.info() in docker_client.py"

just add a TypedDict or something so its not just totally exposing another library API and I think that should be ok

will take a look time permitting

cc @kiview

core/testcontainers/core/config.py
vemonet: Can't they just use the environment variable to configure it if someone ever needs it? A [quick search on GitHub](https://github.com/search?q=TestcontainersConfiguration%28+language%3APython&type=code&l=Python) shows no-one is actually calling `TestcontainersConfiguration` with custom params

Otherwise we can keep `c.ryuk_docker_socket` as a direct property for people to configure

And rename the new computed property to something else, e.g. `c.docker_socket` or `c.get_docker_socket()` (and use this computed property when needing to provide the docker socket)

alexanderankin: see #531

vemonet: I don't really see where is the problem, it is possible to monkeypatch a computed property:

```python
from testcontainers.core.config import testcontainers_config, TestcontainersConfiguration

def custom_ryuk_docker_socket(self) -> str:
    return "/custom/path/to/docker.sock"

TestcontainersConfiguration.ryuk_docker_socket = property(custom_ryuk_docker_socket)
print(testcontainers_config.ryuk_docker_socket)
```

Otherwise I proposed another solution with a static property + a computed one, would this fit?

CarliJoy:
```python
```python
import docker

client = docker.from_env()
print(client.api.get_adapter(client.api.base_url).socket_path)
```

Could be used to detect the correct socket_path.

If `socket_path` is not defined, just use the default.

alexanderankin:
closed in favor of #779

Issue #709, opened 2024-10-02
Bug:  `get_exposed_port()` waits forever, but is fixed by starting the container twice


vemonet:
**Describe the bug**

Using `container.get_exposed_port()` waits forever. Adding an additional `container.start()` starts a 2nd container, but somehow fixes the issue

**To Reproduce**

```python
with DockerContainer(TRIPLESTORE_IMAGE) as container:
    container.with_exposed_ports(7200)
    delay = wait_for_logs(container, "Started GraphDB")
    print("DELAY OK", delay)
    host = container.get_container_host_ip()
    print("HOST OK", host)
    port = container.get_exposed_port(7200)
    # STUCK HERE
    print("PORT OK", port)
    base_url = f"http://{host}:{port}"
    print(f"GraphDB started in {delay:.0f}s at {base_url}")
```

Output (wait forever after last line):

```
tests/test_graphdb.py using host unix:///run/user/1000/docker.sock
using host unix:///run/user/1000/docker.sock
Pulling image testcontainers/ryuk:0.8.1
Container started: 158a1102d52d
Waiting for container <Container: 158a1102d52d> with image testcontainers/ryuk:0.8.1 to be ready ...
Pulling image ontotext/graphdb:10.7.4
Container started: a478700f1fb6
DELAY OK 12.14159607887268
HOST OK localhost
Waiting for container <Container: a478700f1fb6> with image ontotext/graphdb:10.7.4 to be ready ...
```

Funny dirty fix: you can make it work by adding a `container.start()`. It starts a second container, but for some reason the `wait_for_logs` then behave as expected (take ~16s for me)

```python
from testcontainers.core.container import DockerContainer
from testcontainers.core.waiting_utils import wait_for_logs

with DockerContainer("ontotext/graphdb:10.7.4") as container:
    container.with_exposed_ports(7200)
    container.start()
    delay = wait_for_logs(container, "Started GraphDB")
    base_url = f"http://{container.get_container_host_ip()}:{container.get_exposed_port(7200)}"
    print(f"GraphDB started in {delay:.0f}s at {base_url}")
```

Output:

```
tests/test_graphdb.py using host unix:///run/user/1000/docker.sock
using host unix:///run/user/1000/docker.sock
Pulling image testcontainers/ryuk:0.8.1
Container started: 5c72a83ac187
Waiting for container <Container: 5c72a83ac187> with image testcontainers/ryuk:0.8.1 to be ready ...
[...]

PR #711, opened 2024-10-03
docs: Add an example for configuring a custom DockerContainer with exposed port and environment variale

vemonet:
Current examples for running a `DockerContainer` are all using `with DockerContainer` (3 examples out of 3). 

But this is not the right way when someone wants to add some custom configuration (expose a port, define an environment variable), which is quite a regular use-case.

Newcomers reading this docs might get confused and will need to go through the process of finding out how to do this properly outside of the docs (e.g. this guy here https://github.com/testcontainers/testcontainers-python/issues/709 )

So I have added a basic example, feel free to change the docker image/port for something else

### DIFF
--- a/core/README.rst
+++ b/core/README.rst
@@ -30,3 +30,18 @@ Using `DockerContainer` and `DockerImage` to create a container:
 
 The `DockerImage` class is used to build the image from the specified path and tag.
 The `DockerContainer` class is then used to create a container from the image.
+
+
+Creating a container with an exposed port and a custom environment variable:
+
+.. doctest::
+
+    >>> from testcontainers.core.container import DockerContainer
+    >>> from testcontainers.core.waiting_utils import wait_for_logs
+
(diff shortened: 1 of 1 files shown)


Yourgene:
Hi, I had the same "issue" as you few days ago, and I had to look into code to find out that DockerContainer class is much rich than what is displayed in docs.
Thanks for adding it!

If I had a suggestion, I'd also add the volume mapping example, just for the sake of it. Volume mapping is a basic use case too, as the ones you provided.
If needed, here's the line you could add:
```
.with_volume_mapping('/path/to/host/dir/', "path/to/container/dir")
```

Tranquility2:
Looks like this needs some more work

vemonet:
@Tranquility2 it's litterally 10 lines in a readme of the most basic example of how most people will use this lib

Should not be too hard to get right no? 

Note the one I wrote was almost a year ago, so the lib might have evolved since then

Issue #679, opened 2024-08-13
Feature: dependsOn functionality for DockerContainer


mgorsk1:
**What are you trying to do?**

Improve TC-python feature parity with TC-java by enabling container to specify `dependsOn`, which should work similar to how `depends_on` in docker-compose does - container B that depends on container A won't be started until container A is in `running state`. 

**Why should it be done this way?**

Describe the motivation of the enhancement.

**Other references:**

Include any other relevant reading material about the enhancement.

hari134:
I have implemented the depends_on functionality for DockerContainer, mirroring the dependsOn behavior in TC-java so that dependent containers only start once their dependencies are in the running state. The PR for this enhancement is now open: #728. Please check it out, and let me know if any changes are needed.

PR #728, opened 2024-10-26
feat : Add support for depends_on functionality 

hari134:
### Description:
This PR addresses issue #679 by introducing the depends_on functionality to DockerContainer, enhancing feature parity with the dependsOn functionality in TC-java by allowing containers to define dependencies. With this feature, a container can now specify dependencies on other containers, ensuring they will only start once all dependencies reach the running state.

### Changes

- Added a depends_on method to define dependencies between containers.

- Implemented _start_dependencies to manage recursive startup and ensure each dependency reaches the running state.

- Integrated exception handling for cleanup, ensuring that all dependencies are stopped if a failure occurs during the startup process.

### Sample Usage
```python

db_container = DockerContainer("mysql:latest").with_command("tail -f /dev/null")
app_container = DockerContainer("myapp:latest").with_command("tail -f /dev/null").depends_on(db_container)

app_container.start()

app_container.stop()
db_container.stop()

```

### Added test cases to verify:
- Complete cleanup when all dependencies fail to start.

- Cleanup of all dependencies when some dependencies succeed, and others fail.

- Ensuring the main container does not start if any dependency fails.

### DIFF
--- a/core/testcontainers/core/container.py
+++ b/core/testcontainers/core/container.py
@@ -1,9 +1,11 @@
 import contextlib
+import time
 from socket import socket
 from typing import TYPE_CHECKING, Optional, Union
 
 import docker.errors
 from docker import version
+from docker.errors import NotFound
 from docker.types import EndpointConfig
 from typing_extensions import Self, assert_never
 
@@ -44,6 +46,7 @@ class DockerContainer:
         self.env = {}
         self.ports = {}
         self.volumes = {}
+        self._dependencies = []
         self.image = image
         self._docker = DockerClient(**(docker_client_kw or {}))
         self._container = None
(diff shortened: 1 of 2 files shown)


hari134:
Any update on this?

core/testcontainers/core/container.py
```diff
@@ -83,14 +86,63 @@ def maybe_emulate_amd64(self) -> Self:
...
+                    dependency._start_dependencies()
+                    dependency.start()
+                    started_dependencies.append(dependency)
+
+                    if not dependency.wait_until_running():
+                        raise ContainerStartException(f"Dependency {dependency.image} did not reach 'running' state.")
+
+                except Exception as e:
+                    # Clean up all dependencies started before the failure
+                    for dep in started_dependencies:
```
alexanderankin, line 126: this doesn't seem like it work as the recursive call would create its own started_dependencies?

alexanderankin:
@hari134 currently the priorities that I have for the project are making the core parts of it work correctly, rather than adding features (as its a high potential to need to be re-done later). I plan on doing another big push to move the project before the end of the year but not sure when. usually more downtime during holiday season, so this may get merged then.

hari134:
@alexanderankin I completely understand the focus on stabilizing the core functionality before adding new features. I'll address the issues raised, and you can review them whenever it's convenient. Thanks.

core/testcontainers/core/container.py
hari134: @alexanderankin I have made the needed changes such that the _start_dependencies function passes the started_dependencies across the recursive function calls to maintain a consistent state. I have also added a new check for circular dependencies which is called while adding dependencies in depends_on function. The tests have also been updated to be more comprehensive. I will be happy to make changes in the future if any of the core parts change just tag me in that case.

hari134:
@alexanderankin is this ready to merge or are there any changes to be made?

PR #735, opened 2024-11-14
fix(trino): add delay time to avoid Trino issue

grieve54706:
I found the trino container usually query failed due to `nodes is empty`.
I tried to check the catalog TPCH is ready and the node is ready. I still encountered the problem.
The final solution is sleeping for a few seconds.

Relate issue
https://github.com/trinodb/trino/discussions/13388

### DIFF
--- a/modules/trino/testcontainers/trino/__init__.py
+++ b/modules/trino/testcontainers/trino/__init__.py
@@ -11,6 +11,7 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 import re
+import time
 
 from testcontainers.core.config import testcontainers_config as c
 from testcontainers.core.generic import DbContainer
@@ -24,12 +25,14 @@ class TrinoContainer(DbContainer):
         image="trinodb/trino:latest",
         user: str = "test",
         port: int = 8080,
+        delay: int = 5,
         **kwargs,
     ):
         super().__init__(image=image, **kwargs)
         self.user = user
         self.port = port
         self.with_exposed_ports(self.port)
+        self.delay = delay
 
     @wait_container_is_ready()
     def _connect(self) -> None:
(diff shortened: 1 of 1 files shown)


alexanderankin:
what exception does it throw? can't that exception be added to wait_container_is_ready or one of the other decorator we have and it will catch that and rerun the waiter method? this module seems to be abusing the decorators a bit here....

grieve54706:
The trino logs in the container
```
2024-11-18 11:46:00 2024-11-18T03:46:00.216Z    INFO    main    io.trino.server.Server  Server startup completed in 7.66s
2024-11-18 11:46:00 2024-11-18T03:46:00.216Z    INFO    main    io.trino.server.Server  ======== SERVER STARTED ========
2024-11-18 11:46:00 2024-11-18T03:46:00.844Z    INFO    dispatcher-query-3      io.trino.event.QueryMonitor     TIMELINE: Query 20241118_034600_00000_g6qa4 :: FINISHED :: elapsed 379ms :: planning 119ms :: waiting 26ms :: scheduling 216ms :: running 32ms :: finishing 12ms :: begin 2024-11-18T03:46:00.425Z :: end 2024-11-18T03:46:00.804Z
2024-11-18 11:46:01 2024-11-18T03:46:01.070Z    INFO    dispatcher-query-5      io.trino.event.QueryMonitor     TIMELINE: Query 20241118_034600_00001_g6qa4 :: FAILED (GENERIC_INTERNAL_ERROR) :: elapsed 197ms :: planning 197ms :: waiting 0ms :: scheduling 0ms :: running 0ms :: finishing 0ms :: begin 2024-11-18T03:46:00.871Z :: end 2024-11-18T03:46:01.068Z
```
You can see the `SERVER STARTED` log; the first query is `FINISHED` but the second is `FAILED`.

But the first query is `SELECT 1`. Maybe I can use `tpch.tiny.nation` and wait for it. WDYT?
```python
while time.time() < deadline:
    try:
        cur = conn.cursor()
        cur.execute("SELECT * FROM tpch.tiny.nation LIMIT 1")
        cur.fetchall()
        return
    except Exception:
        time.sleep(1)
```

grieve54706:
Hi @alexanderankin, could you recheck this PR?

Issue #687, opened 2024-08-17
Feature: Support static env config from file


Tranquility2:
**What are you trying to do?**

Add missing implementation of ```docker run --env-file <file> ...```

**Why should it be done this way?**

Users should not be required to load each env var manually if they have an env file.
(this is the common case for quite a few services)

For example, if I would have an Redis service with some [config](https://github.com/redis-developer/redis-bike-co/blob/main/env.example):
```env
# Redis Bike Commpany Demo Application: Example .env file.
REDIS_URL=redis://localhost:6379/?decode_responses=True
BIKE_INDEX_NAME="idx:bikes"
STORE_INDEX_NAME="idx:stores"
REDIS_KEY_BASE="redisbikeco"
BIKE_KEY_BASE="redisbikeco:bike"
STORE_KEY_BASE="redisbikeco:store"
FLASK_ENV=development
```
We would now be able to load it without evoking `with_env` 7 times + keeping a single source of truth for the env config.

**Other references:**

[Docker official Docs](https://docs.docker.com/reference/cli/docker/container/run/)

biniyam69:
@Tranquility2 wanna work on this together?

Tranquility2:
@biniyam69 checkout https://github.com/testcontainers/testcontainers-python/pull/684 if you have any comment/ideas/remarks I'll be happy to update.

PR #737, opened 2024-11-16
feat(core): Support working with env files

Tranquility2:
Fix: https://github.com/testcontainers/testcontainers-python/issues/687

Users should not be required to load each env var manually if they have an env file.
Added support for loading a dot-env file.

Usage:
```python
with DockerContainer("nginx:latest").with_env_file("my_env_file"):
    # We now have a container with the relevant env vars from the file
```

This is an implementation of ```docker run --env-file <file> ...```


totallyzen (approved):
LGTM, thanks for the conrtrib!

Issue #727, opened 2024-10-25
Bug: hardcoded dialect/driver parameter in connection url in mysql container


nightblure:
**Describe the bug**

I find it strange that the dialect or driver parameter is hardcoded here (look screenshot below), because today there are [many different methods and drivers for connecting to mysql](https://docs.sqlalchemy.org/en/20/dialects/mysql.html#module-sqlalchemy.dialects.mysql.mariadbconnector). If I use something other than pymysql, I get the wrong connection address from the `get_connection_url` method.

Code:
![Image](https://github.com/user-attachments/assets/1d554724-da72-4024-b9e1-075e3cde644a)

**Possible solution**
1. Remove this hardcoded line and leave only `mysql://...`;
2. Allow you to pass an optional dialect or driver parameter to the `get_connection_url` function or the `MySqlContainer` class initializer. If its value is non-empty, then in the `get_connection_url` method it will be inserted into the right place in the connection string

---

**To Reproduce**

- just see to current code (testcontainers 4.8.2)

**Runtime environment**
Looks optional for this issue

nightblure:
If this solution works, I could do it

totallyzen:
Please feel free to contribute!

I would recommend putting the dialect in the constructor, not `self.get_connection_url()` - it's where the other settings go as well by the looks of it.

PR #739, opened 2024-11-18
fix(mysql): add dialect parameter instead of hardcoded mysql dialect

nightblure:
closes https://github.com/testcontainers/testcontainers-python/issues/727

* add parameter `dialect`;
* tests fixing and add some assertions


nightblure:
@totallyzen hi!

please help me figure out why core tests fail when building a package with this error:

```
Run poetry build && poetry run twine check dist/*.tar.gz
Building testcontainers (4.8.2)
  - Building sdist
  - Built testcontainers-4.8.2.tar.gz
  - Building wheel
  - Built testcontainers-4.8.2-py3-none-any.whl
  File "/home/runner/.cache/pypoetry/virtualenvs/testcontainers-Wc8dTDd--py3.10/bin/twine", line 5, in <module>
    from twine.__main__ import main
  File "/home/runner/.cache/pypoetry/virtualenvs/testcontainers-Wc8dTDd--py3.10/lib/python3.10/site-packages/twine/__init__.py", line 43, in <module>
    __license__ = metadata["license"]
  File "/home/runner/.cache/pypoetry/virtualenvs/testcontainers-Wc8dTDd--py3.10/lib/python3.10/site-packages/importlib_metadata/_adapters.py", line 54, in __getitem__
    raise KeyError(item)
KeyError: 'license'
```

I would like to fix this, but I don’t understand what’s wrong here

modules/mysql/testcontainers/mysql/__init__.py
```diff
@@ -104,8 +106,9 @@ def _connect(self) -> None:
         )
 
     def get_connection_url(self) -> str:
+        dialect = "mysql" if self.dialect is None else f"mysql+{self.dialect}"
```
totallyzen, line 109: I think you should put this simply in the `init` when you set `self.dialect`. 

It would effectively become:
```python
self.dialect = dialect or environ.get("MYSQL_DIALECT") or "mysql"
```

also I think `dialect="mysql+asdf"` makes more sense than `dialect="asdf"`

nightblure: I think that dialect is a specific extension that comes after "mysql". Why repeat “mysql” everywhere if we already know that we are connecting to mysql?

nightblure: @totallyzen Not that this is an argument, but it was interesting to look at the [current implementation of the postgresql container](https://github.com/testcontainers/testcontainers-python/blob/main/modules/postgres/testcontainers/postgres/__init__.py#L55), and it is done according to the principle that I proposed

balint-backmaker: > and it is done according to the principle that I proposed. what do you think?

sure I don't mind - maybe then validate the dialect _does not_ start with `mysql` (raise a `ValueError`) then I'm happy :)

nightblure: ready to make this check what proposed @balint-backmaker if you approve it @totallyzen

totallyzen: oops, sorry for the mix - commented from the company account and not my personal :sweat_smile:

nightblure: @totallyzen did it in last commit

nightblure: @totallyzen ping)

totallyzen:
hi! sorry for the delays, I'll do one more review and merge it for you.

totallyzen (approved):
small window for error: `MYSQL_DIALECT` can still be `mysql+something` and it will break tests. But I think it's good enough like this. I'll merge and release this for you and you can raise another patch for the bug above

Issue #631, opened 2024-07-01
RFC: v5


alexanderankin:
my v5 wishlist:

1. typing/`py.typed` (finally)
1. oop encapsulate things like waiting and such (finally)
1. how dind works - i think after typing and oop encapsulation this one should be much easier to work on.

smaller things:

1. the logger setup as mentioned in https://github.com/testcontainers/testcontainers-python/issues/600#issuecomment-2151451091
1. module split / generic module (DbContainer, others)
1. review most function names - lot of unnecessary variety here

PR #740, opened 2024-11-23
fix(mosquitto): module name

crankcase:
Mosquitto module used wrong name mqtt, but mqtt its name of protocol and also other modules use naming convention `<module name>.<ModuleName>Container`

There also wrong name in community-module-registry documentation https://github.com/testcontainers/community-module-registry/blob/main/modules/mosquitto/index.md?plain=1#L16 

I have prepared separate RP https://github.com/testcontainers/community-module-registry/pull/97


alexanderankin:
I'm not sure we want to do this as it would be a breaking change. maybe a v5 item? https://github.com/testcontainers/testcontainers-python/issues/631

PR #750, opened 2024-12-15
fix(tests): replace dind-test direct docker usage with sdk

### LINKED ISSUE #749: TODO: add a dev note regarding sudoless docker
Devs should be informed (using the official docs) that the testing environment considers `sudoless docker` as a baseline (and some things might not work / break if sudo is required) 

For a future plan, we should rewrite / avoid using docker command directly (for example using `subprocess.run`) and use the docker SDK.

Tranquility2:
fix: #749

Replace using `subprocess.run` in the **dind** test with SDK to avoid the direct call to docker that enforces the developer to run _sudoless_ docker.

### DIFF
--- a/core/tests/conftest.py
+++ b/core/tests/conftest.py
@@ -2,7 +2,6 @@ from pathlib import Path
 
 import pytest
 from typing import Callable
-import subprocess
 from testcontainers.core.container import DockerClient
 import sys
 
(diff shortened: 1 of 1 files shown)


Tranquility2:
Not sure this is the best solution, I'm also considering using the image API to be more verbose.

alexanderankin (approved):

PR #758, opened 2025-01-14
Do not setup log handlers to write to stderr

draftcode:
By setting up a handler in the library, the library users do not have a good control on how/where to write the logs. Do not set a handler from the library.

### DIFF
--- a/core/testcontainers/core/utils.py
+++ b/core/testcontainers/core/utils.py
@@ -14,9 +14,6 @@ WIN = "win"
 def setup_logger(name: str) -> logging.Logger:
     logger = logging.getLogger(name)
     logger.setLevel(logging.INFO)
-    handler = logging.StreamHandler()
-    handler.setLevel(logging.INFO)
-    logger.addHandler(handler)
     return logger
 
 


alexanderankin:
we should add docs and do a logging overhaul i think - e.g. https://github.com/testcontainers/testcontainers-java/blob/2707f3143d3cfa8351f727bfd5752c1155818bd6/core/src/main/java/org/testcontainers/containers/GenericContainer.java#L670-L672

CarliJoy:
@alexanderankin what do you mean with a logging overhaul?

Logging is always quite language specific as it requires global objects.

I am with draftcode that a libary should not setup a logger, it only should log.

I would even go that far, that the library should not even define the log level. It up to the user of the lib to define it.
But I see it is convenient to this. But at least  do it in away, that is done exactly once during import for all modules.

I.e. 

```python

import logging
from typing import Final

LOGGER: Final = logging.getLogger("testcontainers")
LOGGER.setLevel(logging.INFO)
```

This would set the log level to all `testcontainers` logger (i.e. also `testcontainers.core.utils`).

Besides this, there is nothing more to do.

### Why this is important
I am using testcontainers in pytest plugin, as I need containers to be started before imports are done (legacy code, that setups databases during import...).
And there is a [long standing issues with pytest](https://github.com/pytest-dev/pytest/issues/5743) that doesn't handle it well when prints are happening in certain conditions.

So being able to *simply* disable/change the handler all together would be nice.

rhoban13:
Agree 💯 here.  I'd go so far as to suggest `setup_logger` be removed from the library.   All callers should be replaced with simply `logging.getLogger`

alexanderankin:
closed in favor of #836

PR #771, opened 2025-02-12
fix(core): create_label test

neykov:
Make sure the test covers the intended behaviour. Previously it was executing the create_labels function without actually verifying there are no side effects.


neykov:
FYI @alexanderankin this is a follow up to https://github.com/testcontainers/testcontainers-python/pull/769, to actually make the test useful.

alexanderankin (approved):

Issue #786, opened 2025-03-15
Security-Issue: tj-actions/changed-files Action


Tranquility2:
@realjenius Updated:
> Hello all! Out of an abundance of caution [due to this security report](https://github.com/tj-actions/changed-files/issues/2463), Docker Security has disabled Github Actions for the testcontainers-python repository until we can remove use of the tj-actions/changed-files Github Action in the workflows.
> 
> The suggested action is to move to an alternative solution (like [path-filter](https://github.com/dorny/paths-filter)) with no known vulnerabilities.  Sorry for the inconvenience!

alexanderankin:
https://github.com/tj-actions/changed-files

alexanderankin:
GitHub simply removed this repository?

Tranquility2:
That's one option or blocked access, "Someone committed a base64-encoded payload that runs a script that in turn prints out encoded secrets…"

realjenius:
Thank you all - appreciate everyone taking care of this so quickly! 👍

PR #787, opened 2025-03-15
fix(security): Update track-modules job 

Tranquility2:
This will address the Security issue reported on #786
As recommended, `tj-actions/changed-files` was replace replaced with [path-filter](https://github.com/dorny/paths-filter)

### DIFF
--- a/.github/workflows/ci-community.yml
+++ b/.github/workflows/ci-community.yml
@@ -22,7 +22,7 @@ jobs:
           fetch-depth: 0
       - name: Get changed files
         id: changed-files
-        uses: dorny/paths-filter@v3
+        uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36 # v3
         with:
           base: ${{ github.ref }}
           list-files: 'json'
--- a/.github/workflows/ci-community.yml
+++ b/.github/workflows/ci-community.yml
@@ -19,20 +19,20 @@ jobs:
       - name: Checkout contents
         uses: actions/checkout@v4
         with:
-          fetch-depth: 0  # recommended by tj-actions/changed-files
+          fetch-depth: 0
       - name: Get changed files
         id: changed-files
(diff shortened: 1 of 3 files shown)


Tranquility2:
Note to reviewer, I keep the testing so we can see it works the same.
Testing locally is very limited (I used [act](https://github.com/nektos/act))
the final result looks like
```[modules/track-modules]   ⚙  ::set-output:: computed_modules=["aws","generic"]```
which looks compatibale with what we had on (for example) https://github.com/testcontainers/testcontainers-python/actions/runs/13502155079/job/37723003600?pr=778
image

```computed_modules=["scylla"]```

File: .github/workflows/ci-community.yml [hunk 1]
```diff
@@ -19,20 +19,20 @@ jobs:
...
         id: changed-files
-        uses: tj-actions/changed-files@v42
+        uses: dorny/paths-filter@v3
         with:
-          path: "./modules"
-          diff_relative: true
-          dir_names: true
-          dir_names_exclude_current_dir: true
-          json: true
+          base: ${{ github.ref }}
```
Tranquility2, line 27: Note: this was needed locally, we can try without on a later stage

.github/workflows/ci-community.yml
```diff
@@ -19,20 +19,20 @@ jobs:
       - name: Checkout contents
         uses: actions/checkout@v4
         with:
-          fetch-depth: 0  # recommended by tj-actions/changed-files
+          fetch-depth: 0
       - name: Get changed files
         id: changed-files
-        uses: tj-actions/changed-files@v42
+        uses: dorny/paths-filter@v3
```
mdelapenya, line 25: suggestion: Can we use the pinned version of v3 (passing the commit hash)?

mdelapenya (commented):
BTW the GH action was restored removing the offending commit. In the case we want to keep it, I'd encourage using a pinned version of the action. And if possible, also pin any other GH action

PR #793, opened 2025-03-27
feat(modules): add Azure Service Bus Emulator module

### LINKED ISSUE #792: New Container: Azure Service Bus Emulator

[Azure Service Bus Emulator](https://github.com/Azure/azure-service-bus-emulator-installer), this exists within the Java and C# implementations and it would be good to have feature parity.

The container has a MSSQL container dependency to run along with some annoying EULA acceptance features. Mounting configuration into the container is also required rather than configuration through the environment. It doesn't help that the emulator implementation isn't open source.

**Other references:**

Include any other relevant reading material about the enhancement.

Poogles:
Resolves #792.


modules/azure/testcontainers/azure/_types.py
Poogles, line 1: Honestly I'm not too convinced on this, I think maybe using `dataclasses.dataclass` might be better? They offer a JSON schema schema [here](https://github.com/Azure/azure-service-bus-emulator-installer/blob/main/ServiceBus-Emulator/Schema/Config-schema.json) that could be used as a source.

Open to thoughts.

PR #810, opened 2025-05-05
fix(core): mypy

alexanderankin:
makes mypy happier


alexanderankin:
here we go

dsully:
When will this be released to Pypi? Thanks!

alexanderankin:
currently looking for beta testers to test main and tell me its ok to release. last time i released, it broke a bunch of people and issue was filed within like 2 hours. not sure if the fix worked.

PR #815, opened 2025-05-18
Use 'latest' tag for Qdrant module 

w1am:
Newer versions of Qdrant include useful features like the [Batch Search API](https://qdrant.tech/documentation/concepts/search/#batch-search-api), which aren't available in older versions.

Instead of hardcoding a specific version, this PR sets the default image tag to `latest` so we automatically benefit from new features. While it's possible to override the image manually, having `latest` as the default keeps things simple, reduces boilerplate, and ensures we're testing against the most up-to-date Qdrant.

### DIFF
--- a/modules/qdrant/testcontainers/qdrant/__init__.py
+++ b/modules/qdrant/testcontainers/qdrant/__init__.py
@@ -39,7 +39,7 @@ class QdrantContainer(DbContainer):
 
     def __init__(
         self,
-        image: str = "qdrant/qdrant:v1.8.3",
+        image: str = "qdrant/qdrant:latest",
         rest_port: int = 6333,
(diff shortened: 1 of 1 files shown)


alexanderankin (approved):

alexanderankin:
unfortunately this is not correctly reflected in the library but using the class without specifying the version is deprecated - for example see - https://github.com/testcontainers/testcontainers-java/blob/7d8301903a6b45591605fd376702d24fc878a61d/modules/postgresql/src/main/java/org/testcontainers/containers/PostgreSQLContainer.java#L44-L50 - i do not think having unpredictable breakages on a major version upgrade is something people want but i will update it this time to alleviate the pain point here

PR #836, opened 2025-06-18
fix: just use the getLogger API and do not override logger settings

alexanderankin:


alexanderankin:
@Tranquility2 @balint-backmaker @santi @kiview help! please merge

Tranquility2 (approved):

PR #842, opened 2025-07-02
fix: Enable mypy in the CI

Tranquility2:
1. leftovers from a previous mypy refactor
```python
core/testcontainers/core/config.py:42: error: Redundant cast to "str"  [redundant-cast]
            socket_path = cast("str", socket_path)
 ```
 
 Post fix:
 ```bash
#poetry run mypy --config-file pyproject.toml core
Success: no issues found in 36 source files
```

2. add mypy check to the CI under lint, also create a simple report if something is wrong.


Tranquility2:
Happy flow:
image

On error:
image

PR #852, opened 2025-07-28
feat: Transferable

rhoban13:
In https://github.com/testcontainers/testcontainers-python/pull/676 for copying to/from `DockerContainer`, it was suggested that we should clarify the interface & usage a bit before proceeding.  This PR aims to push that conversation forward with some test cases illustrating proposed usages.  Although here I created a `Transferable` object, in most cases there's no need for the caller to explicitly construct a Transferable, just pass the `bytes | Path`

**Proposed use cases:**

1. Copy into a container by passing a `TransferSpec` into the initializer: 
```
DockerContainer(... transferrables=[TransferSpec(source, destination_in_container), ...)
```

2. Copy into the container via the builder pattern, prior to starting the container:
```
DockerContainer(...)
     .with_copy_into_container(b"some_bytes", destination_in_container)
     .with_copy_into_container(some_path, destination_in_container)
 ```

3. Copy into the container at runtime:
```
with DockerContainer(...) as container:
    container.copy_into_container(b"some_bytes", destination_in_container)
    container.copy_into_container(some_path, destination_in_container)
```

### CODE REVIEW

**Comment by wtf-is-flying** [comment]
Hi @rhoban13, what is the current status of this feature? What could I do to help?

For what it's worth, I think the suggested interface is clear. I haven't tested it with my use cases yet though.

**Comment by rhoban13** [comment]
Tagging a few maintainers seeking feedback @terry-docker @alexanderankin @Tranquility2

**Comment by alexanderankin** [comment]
i think my current priorities are docs, anything that makes this library behave like the others (this PR is illustrating current state not improving it), so far the only new issue from the new release is someone complaining about warning logs - so i think that means that the new release was successful (either that or there are no more users of this library). anecdotally someone reached out on slack asking for a new release but they are only allowed to use it at work a week after it has been released. so there may be new issues in some time.

**Comment by alexanderankin** [comment]
wait, this adds Transferable - this is great - if we can also fix the immediate starting of the container this almost entirely fixes copying. just have to maybe see about creating folders automatically and i think we may have parity on copying files with the other impls

**Comment by rhoban13** [comment]
Yes I was just about to update the PR title - my original intent was just some tests, but I found it easier to write those tests using TDD & just implement it at the same time.  I think this brings the library a bit closer to feature parity with some others.

I'll take a look at adding some tests around folders if it sounds like you're in agreement with the interface at a high level.

**Comment by wtf-is-flying** [comment]
Thanks! I'll try to run some tests then and will let you know how they went.

**Comment by rhoban13** [comment]
> if we can also fix the immediate starting of the container

This is independent of this PR, right?  I presume you're referring to the fact that we're calling  [docker run](https://github.com/testcontainers/testcontainers-python/blob/aa474359d4e8d2c77e2c195352261d424ee5ca7f/core/testcontainers/core/container.py#L195) instead of create+start?

**Comment by alexanderankin** [comment]
yeah this whole API is not going to behave as intended until we start doing create then start

**Comment by wtf-is-flying** [comment]
I ran some tests with this PR, it went well 🙂.

I used this function to transfer directories:

```python
def transferables_from_directory(
    source_dir: Path, dest_dir: Path, mode: int
) -> list[TransferSpec]:
    transferables: list[TransferSpec] = []
    for dirpath, _, filenames in source_dir.walk():
        for name in filenames:
            origin = dirpath / name
            destination = dest_dir / dirpath.relative_to(source_dir) / name
            transferables.append((origin, str(destination), mode))

    return transferables
```

rhoban13:
I updated the implementation so all these methods should work if the transferable Path is a directory as well.

alexanderankin:
going to ping @Tranquility2 in an attempt to make sure this PR doesnt get forgotten - I want either this or the DB stuff to be the next release (sometime in the next couple weeks)

Tranquility2:
> going to ping @Tranquility2 in an attempt to make sure this PR doesnt get forgotten - I want either this or the DB stuff to be the next release (sometime in the next couple weeks)

How can I assist?

PR #853, opened 2025-08-01
fix(postgres): Use with_bind_ports for postgres container

hannah-heywa:
with_exposed_ports for ports != 5432 breaks -- the internal port is always 5432, we're just binding to different host ports.


alexanderankin:
what are you trying to do and what is happening instead? testcontainers generates a random port - this behavior is ideal for tests, so that you can run tests in parallel, or even you can run the tests on a system where postgres is running on the default port (a random port wont conflict). that is the intended behavior.

if you can share some details about what you are doing, i can help you figure out how to use testcontainers in your workflow.

PR #863, opened 2025-08-19
fix(core): improper reading of .testcontainers.properties

### LINKED ISSUE #864: Bug: .testcontainers.properties file not properly used
**Describe the bug**

When specifiying ryuk.disabled and ryuk.containers.privileged, the values are not listened to in ~/testcontainers.properties

**To Reproduce**

Set
```
ryuk.disabled=true
ryuk.container.privileged=false
```
in ~/.testcontainers.properties

**Runtime environment**

Irrelevant. The code for handing containers is written incorrectly.

https://github.com/testcontainers/testcontainers-python/pull/863

laundry-96:
fix #864 

The environment variables were not overridden from the .testcontainers.properties file for ryuk variables. This causes the properties file to never actually be used. This commit detects the environment variable, and if unspecified falls back to the properties file, and if not specified, defaults to false

### DIFF
--- a/core/testcontainers/core/config.py
+++ b/core/testcontainers/core/config.py
@@ -11,6 +11,8 @@ from typing import Final, Optional, Union
 
 import docker
 
+ENABLE_FLAGS = ("yes", "true", "t", "y", "1")
+
 
 class ConnectionMode(Enum):
     bridge_ip = "bridge_ip"
@@ -52,7 +54,7 @@ def get_bool_env(name: str) -> bool:
     Defaults to False.
     """
     value = environ.get(name, "")
-    return value.lower() in ("yes", "true", "t", "y", "1")
+    return value.lower() in ENABLE_FLAGS
 
 
 TC_FILE = ".testcontainers.properties"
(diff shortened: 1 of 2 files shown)


alexanderankin:
see if you can make the tests pass? i can take a look to see if i can finish this pr when i am available

laundry-96:
@alexanderankin I fixed it by adding a Optional[bool] defaulted to None, which allows an override for ryuk_disabled and ryuk_privileged programmatically.

Sorry for not reading the CONTRIBUTING.md. I didn't see it in the top level directory and didn't see the README.md 😅 

I'm on a mac, so the test is set to skipped, but I disabled that and it seems to have passed. I can't be too sure until I see the tests here though.

laundry-96:
@alexanderankin Can I get one more run please? I have added some extra test cases and caught some errors myself :)

laundry-96:
@alexanderankin It seems like the docs build failed due to a connection failure? Mind triggering again?

alexanderankin (approved):

alexanderankin:
can you tldr what this code does? im on a work trip right now and just dont have time this week to look at this - i would merge otherwise

laundry-96:
TLDR: Originally when running test containers, you can set variables inside of `~/.testcontainers.properties` and it should be picked up to look at settings, if the corresponding environment variables are not set.

What was happening originally was that the `ryuk_privileged` and `ryuk_disabled` would read the environment, and if it wasn't set, it would default to False. And because the variable is set, it doesn't actually inherit the properties file.

Function causing this issue:
```
def get_bool_env(name: str) -> bool:
    """
    Get environment variable named `name` and convert it to bool.
    Defaults to False.
    """
    value = environ.get(name, "")
    return value.lower() in ("yes", "true", "t", "y", "1")
 ```
    
As you can see this would return either True or False.

What my change does, is bring this function into the class, and when attempting to access the disabled or privileged class value, it calculates it based on if it was set previously (via code, env, or properties file), and if not then checks the environment, then checks the properties file, then defaults to False

Also, I totally understand being busy! I just want to say thank you so much for providing this package from your own time and energy :)

PR #876, opened 2025-09-22
fix(core): make sure context manager exits

alexanderankin:
maybe don't write your software in python


santi:
Looked through it, and my initial thought was that this is somewhat similar to the problem that [ExitStack](https://docs.python.org/3/library/contextlib.html#replacing-any-use-of-try-finally-and-flag-variables) tries to solve, but at the same time ExitStack is very niche, so that is it much more maintainable to do it your way

alexanderankin:
exit stack takes so much more code. its kind of horrific either way. im open to improvements now or later.

santi (approved):

PR #883, opened 2025-09-30
fix(modules): Fix postgres deprecation warning.

surister:
Removed the `wait_container_is_ready` decorator as per deprecation warning into the new `container.waiting_for`

### DIFF
--- a/modules/postgres/testcontainers/postgres/__init__.py
+++ b/modules/postgres/testcontainers/postgres/__init__.py
@@ -64,6 +64,11 @@ class PostgresContainer(DbContainer):
         self.driver = f"+{driver}" if driver else ""
 
         self.with_exposed_ports(self.port)
+        self.waiting_for(
+            LogMessageWaitStrategy(
+                'database system is ready to accept connections').with_startup_timeout(10)
+        )
+
 
     def _configure(self) -> None:
         self.with_env("POSTGRES_USER", self.username)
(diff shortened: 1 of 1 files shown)


surister:
The string `database system is ready to accept connections` is valid for all postgres versions from postgres:9 to postgres:18

surister:
I was testing this thoroughly and sometimes, maybe 1/10 runs this fails, the log successfully emits 'database system is ready...' and the waiting strategy picks it up but still the database is not ready. 

```log
<re.Match object; span=(1342, 1388), match='database system is ready to accept connections'>
... we try connect in _connect
ExecResult(exit_code=2, output=b'psql: error: connection to server at "127.0.0.1", port 5432 failed: Connection refused\n\tIs the server running on that host and accepting TCP/IP connections?\n')
``` 

Adding a `1s` sleep in the `_connect` makes this go away, the funny thing is that postgres says it's ready when it's actually not.

surister:
Pipeline failing due to missing import (bad cherrypicking on my part :smiling_face_with_tear:) Anyway, will close this and properly migrate the module once #892 is finished.

jankatins:
I guess this could be done with a healthcheck waiting strategy or by adding a new waiting strategy which runs a function or runs a command with sh in the container

alexanderankin:
for postgres we disagree with at least java but several people have come to
adjust specifically postgres wait logic so im inclined to preserve behavior
if possible

Issue #884, opened 2025-09-30
What class to use to implement or migrate database modules


surister:
Most modules use the `testcontainers.core.generic.DbContainer` class, but it is marked as deprecated, should we use `testcontainers.generic.ServerContainer`? 

`DbContainer` had a couple of generics for most databases like uri creation and sqlalchemy connect in `_connect`,  do we just lose that and every module needs to implement their own or would it be ok to re-implement `DbContainer` from `ServerContainer` to keep the database specific things?

Tranquility2:
`DbContainer` is great but the idea was to move to a more generic hierarchy where we can extend `ServerContainer` to things like `DbContainer`. The old `testcontainers.core.generic` will be removed eventually after a proper replacement been made, something like `modules.generic.testcontainers.generic.db` (I imagine) that users can migrate from `DbContainer` with ease 😊

surister:
Sounds good, is there any issue where we can find the plan for `modules.generic.testcontainers.generic.db`

surister:
I'm not sure that inheriting `DockerContainer -> ServerContainer -> DBContainer` makes sense, `ServerContainer` is just an HTTP specialization of `DockerContainer`, while some database might implement HTTP most don't.

alexanderankin:
i think the deprecation may be pre-mature in that there is no real alternative right now but it does represent the opinion that we do not want to rely on a sql health check or build the urls the same way for all databases. i think both of these remain bad assumptions. if you want to use them for now i think that is fine - esp if it is merged in here because i will clean that up when we get around to it. i dont want to make people install sqlalchemy if they dont have to. if it is useful for cratedb, lets add it as an optional dependency for the "crate" extra (pip install testcontainers[cratedb])

Tranquility2:
`DockerContainer -> ServerContainer -> DBContainer` is not the direction, sorry if that got lost on the way, `DBContainer ` just should not be part of core.. but as a module like `modules.generic.testcontainers.generic.db` and doesn't have to include `ServerContainer` in any way, hope that help to clarify. 
(I'll try to get a draft ready soon)
The idea/philosophy here that core should not include specific modules (even if we add the word "generic" to them 😅), those are just more modules that we can handle and maintain under `modules.generic`

Tranquility2:
Please checkout https://github.com/testcontainers/testcontainers-python/pull/892

surister:
I think that the information provided in this issue is enough to answer the original question, thank you very much!

PR #892, opened 2025-10-02
feat(generic): Reintroducing the generic SQL module

Tranquility2:
Related to #884
Trying to replace the old generic.py in core to a nicer version under the generic module.

`SqlContainer` its not as good (in being generic) as `ServerContainer` but it should allow us to deprecate `core/testcontainers/core/generic.py` with minimal effort for users, it could lead to a more Generic version like `DBContainer` in the future.

Update 1: Refactor to use `SqlConnectWaitStrategy`
Update 2: Now `SqlConnectWaitStrategy` is required and the users can provide `SqlContainer` with any wait strategy.
Update 3: Now utilizes all the latest improvements from `WaitStrategy`

Note: I think the added tests + documentation (provided in this PR) are by themselves a great improvement over the current generic.py

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -304,9 +304,6 @@ source = [
     "*/site-packages/testcontainers",
     "*/dist-packages/testcontainers",
 ]
-omit = [
-    "core/testcontainers/core/generic.py", # Marked for deprecation
-]
 
 [tool.ruff]
 target-version = "py39"


modules/generic/testcontainers/generic/sql.py
```diff
@@ -0,0 +1,176 @@
...
+
+class SqlContainer(DockerContainer):
+    """
+    Generic SQL database container providing common functionality.
+
+    This class can serve as a base for database-specific container implementations.
+    It provides connection management, URL construction, and basic lifecycle methods.
+    """
+
+    @wait_container_is_ready(*ADDITIONAL_TRANSIENT_ERRORS)
```
surister, line 29: `wait_container_is_ready` is deprecated

alexanderankin:
can we add a new extra? like a db (or more accurately, sql) extra? this is assuming we dont want to aim for complete removal.

Even if we do that, i dont like the _connect method on the container - i think connecting is the business of 1) app code 2) sql waiting strategy - so all the db interface has to do is tell someone how to connect - and maybe this can just return the url?

This second point is made further complicated by now you can only use this waiting strategy on container that have the interface implemented. i think maybe this waiting strategy can be parametrized - with information on how to build the url. this removes the need for the container to build its own url (finally). just a thought.

Tranquility2:
- The new extra sound a bit like an overkill and could be a bit confusing for the users don't you think? My goal was to make the transition as seamless as possible but I agree it might be a good time to rethink some things. I just don't want to "branch out" too much, all in all its a very good point but I concerned it will be lost will all the other sql modules 😓
- I totally understand your point regarding `_connect `, I had one iteration where all the logic was under `ConnectWaitStrategy` (with no `_connect ` at all) and I assumed we want backward compatibility with `DbContainer` so it will be easier to move to (so we can remove `generic.py`), let me try and do another version without the connect and `ConnectWaitStrategy` more parametrized as you suggested.

Tranquility2:
New version is ready:
1. no more `_connect`
2. all extended logic moved to `SqlConnectWaitStrategy` (new file) making `sql.py` much cleaner and more generic
3. smaller improvements to make the code simpler and easier to maintain 

You can now say `SqlConnectWaitStrategy` is basically `sqlalchemyConnectWaitStrategy` 😂 but this can also be improved and made more generic.

surister:
I like the new `SqlConnectWaitStrategy` a lot.  Indirectly the `_connect` method paired with `wait_container_is_ready` did just that in many database modules.

Tranquility2:
Make one more critical improvement, now `wait_strategy` is a param for the `SqlContainer`, this
1. keeps it very generic - all the others can use it with ease
2. provides `SqlConnectWaitStrategy` if anyone needs its
3. allows each module to provide a custom  wait_strategy (that can be as simple as waiting for some log string from the server)

Tranquility2:
@alexanderankin  now using `WaitStrategy._poll` and `with_transient_exceptions` from your latest improvements.
This makes `SqlConnectWaitStrategy` much simpler cleaner, thanks 🙏

modules/generic/testcontainers/generic/sql.py
```diff
@@ -0,0 +1,153 @@
...
+
+        Args:
+            image: Docker image name
+            wait_strategy: Wait strategy for SQL database connectivity
+            **kwargs: Additional arguments passed to DockerContainer
+        """
+        super().__init__(image, **kwargs)
+        self.wait_strategy = wait_strategy
+
+    def _create_connection_url(
```
surister, line 35: Big Saturday nitpick:

I think both `_create_connection_url` and `create_connection_url` can be simplified further.

Nit: I'd call it `URI` not `URL`, as most database connection strings follow https://datatracker.ietf.org/doc/html/rfc3986. But that might break more things.

`_create_connection_url` is the typical method that could be just a function in the module, or a static one, it shouldn't access state from the container, it should have one job, create a valid `uri (or url)` from the given parameters. And `get_connection_url` would be a method that dictates how a given module, e.g. Postgres, gets its parameter (either from current container state or user-given) and modifies those as needed, for example appending '+psycopg2' to the `scheme`.

Now, as postgres (and most databases) follow typical URIs creation, a simple implementation could be implemented that would fit most use cases.

```python
# Pseudocode
class SqlContainer(DockerContainer):
    def _create_connection_url(self, dialect, username, password... etc):
        # Does not access `self` state, could be a simple function in the generics module.
        ...
        return f"{dialect}://{username}:{password}@..." 

    def get_connection_url(self, driver, dialect, username, password... etc):
       dialect = dialect or f`{dialect}+{driver}`
       username = username or self.username
       password = password or self.password
       ...
        return self._create_connection_url(driver, dialect, username, password...)
```
Something like this, this way the user can easily modify those two behaviors, how to parameters and how they are used to create a connection uri.

surister: Note that by doing it this way we are somewhat tightly coupling `SqlContainer` to `sqlalchemy` (due to the dialect+driver part) which should be mostly fine as most if not all sql databases support it. Still it'd trivial to override these parts if properly split as recommended.

surister (approved):
Tried to migrate Postgres and CrateDB module and it's fairly easy, I left a few nitpicks that you might use as inspiration, otherwise LGTM and I'd merge.

modules/generic/testcontainers/generic/sql.py
Tranquility2: Please note that `SqlAlchemyWaitingStrategy` is not the default, just an option

modules/generic/testcontainers/generic/sql.py
```diff
@@ -0,0 +1,153 @@
...
+        Raises:
+            ValueError: If unexpected arguments are provided or required parameters are missing
+            ContainerStartException: If container is not started
+        """
+
+        if self._container is None:
+            raise ContainerStartException("Container has not been started")
+
+        # Validate required parameters
+        if not dialect:
```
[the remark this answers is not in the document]
Tranquility2: Explicate over implicate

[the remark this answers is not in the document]
Tranquility2: Done

modules/generic/testcontainers/generic/sql.py
```diff
@@ -0,0 +1,139 @@
...
+        return self
+
+    def _configure(self) -> None:
+        """
+        Configure the database container before starting.
+
+        Raises:
+            NotImplementedError: Must be implemented by subclasses
+        """
+        raise NotImplementedError("Subclasses must implement _configure()")
```
rhoban13, line 118: I'm curious why make this required to override?   Feels a little interesting for this to be not abstract on the base `Container` and abstract on this derived class.  All the existing modules which implement this seem to just be setting env vars on the container, which is somewhat common, but could also be achieved by passing `env_vars` into the initializer.

modules/generic/testcontainers/generic/providers/sql_connector.py
```diff
@@ -0,0 +1,48 @@
...
+
+import logging
+
+from testcontainers.core.waiting_utils import WaitStrategy, WaitStrategyTarget
+
+logger = logging.getLogger(__name__)
+
+ADDITIONAL_TRANSIENT_ERRORS = []
+try:
+    from sqlalchemy.exc import DBAPIError
```
rhoban13, line 12: Should we consider moving this check into the initializer below?  

If `sqlalchemy` is unavailable, at best a caller could construct an instance of the below class, and it'll raise when `wait_until_ready` is eventually called.  Unless I'm misunderstanding - there doesn't seem to be any value to constructing an instance where `ADDITIONAL_TRANSIENT_ERRORS` is empty?

alexanderankin: ideally one of the benefits of moving this to its own corner (so it can sit and think about what it did) is that sqlalchemy is no longer conditionally available (it either is in the case of the extra being installed, or no)

modules/generic/testcontainers/generic/providers/sql_connector.py
```diff
@@ -0,0 +1,48 @@
...
+
+    ADDITIONAL_TRANSIENT_ERRORS.append(DBAPIError)
+except ImportError:
+    logger.debug("SQLAlchemy not available, skipping DBAPIError handling")
+
+
+class SqlConnectWaitStrategy(WaitStrategy):
+    """Wait strategy for database connectivity testing using SQLAlchemy."""
```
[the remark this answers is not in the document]
Tranquility2: Sure, done

modules/generic/testcontainers/generic/sql.py
```diff
@@ -0,0 +1,139 @@
...
+        """
+        Initialize SqlContainer with optional wait strategy.
+
+        Args:
+            image: Docker image name
+            wait_strategy: Wait strategy for SQL database connectivity
+            **kwargs: Additional arguments passed to DockerContainer
+        """
+        super().__init__(image, **kwargs)
+        self.wait_strategy = wait_strategy
```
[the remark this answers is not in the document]
Tranquility2: on the base its Optional, here its not

PR #896, opened 2025-10-05
feat(core): Add `run_containers` function in utils.py to run several containers

surister:
This utils method runs containers on the same context manager. This ensures that the passed containers are run in order and cleaned up.

The main use case is to be able to run code after all containers have started. I need this to create CrateDB clusters.

Practical example: run containers inside a network and clean up the network (network can only be cleaned up after all containers using it are stopped)

```python
network = Network()
network.create()

with run_containers(
        PostgresContainer(network=network),
        PostgresContainer(image='postgres:16', network=network),
) as containers:
    c1, c2 = containers
    conn1 = sqlalchemy.engine.create_engine(c1.get_connection_url()).connect()
    conn2 = sqlalchemy.engine.create_engine(c2.get_connection_url()).connect()

    result1 = conn1.execute(sqlalchemy.text("select version()")).fetchone()
    result2 = conn2.execute(sqlalchemy.text("select version()")).fetchone()

    print(result1, result2, sep='\n')

# The network gets removed only when containers are stopped.
network.remove()
```

### DIFF
--- a/core/testcontainers/core/utils.py
+++ b/core/testcontainers/core/utils.py
@@ -3,8 +3,10 @@ import os
 import platform
 import subprocess
 import sys
+from collections.abc import Generator
+from contextlib import AbstractContextManager, ExitStack, contextmanager
 from pathlib import Path
-from typing import Any, Final, Optional
+from typing import Any, Final, Optional, TypeVar
 
 LINUX = "linux"
 MAC = "mac"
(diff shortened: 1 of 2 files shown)


alexanderankin:
if we add this can we name it consistently with any one of the other implementations, thanks

surister:
> if we add this can we name it consistently with any one of the other implementations, thanks

Hm, could you like any of the other implementations? Can't find them @alexanderankin

rhoban13:
I'm admittedly not sure this should be part of the library, when the caller can achieve the same result explicitly with the same amount of code.

```
with run_containers(
        PostgresContainer(network=network),
        PostgresContainer(image='postgres:16', network=network),
) as containers:
    c1, c2 = containers
```

Becomes

```
with contextlib.ExitStack() as stack:
    c1, c2 = stack.enter_context(container) for container in (
        PostgresContainer(network=network),
        PostgresContainer(image='postgres:16', network=network)
    )
```

alexanderankin:
https://www.javadoc.io/doc/org.testcontainers/testcontainers/1.12.1/org/testcontainers/lifecycle/Startables.html

alexanderankin:
https://java.testcontainers.org/features/advanced_options/#parallel-container-startup

rhoban13:
Ahhhh, I see what you're looking to achieve.  I could definitely see the parallel container startup being a really useful feature.  In it's current form, that's not what this PR proposes.  

I suspect such an implementation and would warrant some discussion on which of python's plethora of ways to achieve parallelism makes most sense here.

Issue #874, opened 2025-09-11
bug: community modules using old decorators print deprecation warnings


DarthMax:
It seems that the 4.13 release introduced a new deprecation warning around the `@wait_container_is_ready` decorator. 

The deprecation warning itself works as expected and the description is comprehensive. 
There is a problem however: importing `DockerContainer` will cause the deprecation warning to appear with the following stack trace:

```
conftest.py:6: in <module>
    from testcontainers.core.container import DockerContainer
project/venv/lib/python3.13/site-packages/testcontainers/core/container.py:21: in <module>
    from testcontainers.core.wait_strategies import LogMessageWaitStrategy
project/venv/lib/python3.13/site-packages/testcontainers/core/wait_strategies.py:41: in <module>
    from .waiting_utils import WaitStrategy, WaitStrategyTarget
project/venv/lib/python3.13/site-packages/testcontainers/core/waiting_utils.py:178: in <module>
    @wait_container_is_ready()
project/venv/lib/python3.13/site-packages/testcontainers/core/waiting_utils.py:117: in wait_container_is_ready
    warnings.warn(
E   DeprecationWarning: The @wait_container_is_ready decorator is deprecated and will be removed in a future version. Use structured wait strategies instead: container.waiting_for(HttpWaitStrategy(8080).for_status_code(200)) or container.waiting_for(LogMessageWaitStrategy('ready'))
```

Our project is configured to treat warnings as errors. As far as I am aware, there is currently nothing we can do on the consumer side to prevent this issues.

As a side note, the documentation still mentions the `wait_for_logs` method: 
https://testcontainers-python.readthedocs.io/en/latest/core/README.html

**To Reproduce**

```python
from testcontainers.core.container import DockerContainer
```

<details><summary>template details</summary>

**Runtime environment**

Provide a summary of your runtime environment. Which operating system, python version, and docker version are you using? What is the version of `testcontainers-python` you are using? You can run the following commands to get the relevant information.

```bash
# Get the operating system information (on a unix os).
$ uname -a
Linux computer 6.16.5-arch1-1 #1 SMP PREEMPT_DYNAMIC Thu, 04 Sep 2025 23:18:13 +0000 x86_64 GNU/Linux
# Get the python version.
$ python --version
Python 3.13.7
# Get the docker version and other docker information.
$ docker info
Client:
 Version:    28.3.3
 Context:    default
 Debug Mode: false
 Plugins:
  compose: Docker Compose (Docker Inc.)
[...]

PR #903, opened 2025-10-14
fix(kafka): Use wait strategy instead of deprecated wait_for_logs

Hugo-C:
Related to https://github.com/testcontainers/testcontainers-python/issues/874.

Since those containers first need to be started to know their exposed ports, we cannot use `DockerContainer._wait_strategy`. We have to keep checking logs manually in their own `start` methods.

Let me know if there is anything to change.


Hugo-C:
#945 Fixed Kafka but not Redpanda, so I adapted the changes accordingly.

Hugo-C:
@alexanderankin if you can take a look as you merged the similar PR.

PR #906, opened 2025-10-15
fix: get_wrapped_container returns container all the time now

### LINKED ISSUE #905: Bug: HealthcheckWaitStrategy returns error with DockerCompose
**Describe the bug**

When applying a 'waiting_for' to a DockerCompose object with a HealthcheckWaitStrategy, it returns an AttributeError: “ComposeContainer” object has no attribute “attrs”.

**To Reproduce**
```python
import os
from testcontainers.compose import DockerCompose
from testcontainers.core.wait_strategies import HealthcheckWaitStrategy

BASE_DIR = os.path.dirname(__file__)

compose = DockerCompose(context=BASE_DIR, compose_file_name = "docker-compose.yml")
compose.waiting_for({"web": HealthcheckWaitStrategy().with_startup_timeout(60)})
compose.start()
```
docker-compose.yml:
```yml
version: '3.8'
services:
  web:
    image: nginx
    healthcheck:
      test: ["CMD", "curl", "-f", "http://localhost"]
      interval: 30s
      timeout: 10s
      retries: 3
      start_period: 40s
```
Traceback:
```bash
  File "<path>/test.py", line 9, in <module>
    compose.start()
  File "testcontainers/compose/compose.py", line 303, in start
    strategy.wait_until_ready(container)
  File "testcontainers/core/wait_strategies.py", line 487, in wait_until_ready
    health = wrapped.attrs.get("State", {}).get("Health", {})
             ^^^^^^^^^^^^^
AttributeError: 'ComposeContainer' object has no attribute 'attrs'
```
**Runtime environment**

```bash
$ uname -a
Linux kali 6.8.0-85-generic #85-Ubuntu SMP PREEMPT_DYNAMIC Thu Sep 18 15:26:59 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
$ python --version
Python 3.12.1
$ poetry show | grep testcontainers
testcontainers                                  4.13.2 
```

alexanderankin:
fix #905

### DIFF
--- a/core/testcontainers/compose/compose.py
+++ b/core/testcontainers/compose/compose.py
@@ -11,6 +11,8 @@ from subprocess import run as subprocess_run
 from types import TracebackType
 from typing import Any, Callable, Literal, Optional, TypeVar, Union, cast
 
+from docker import DockerClient
+from docker.models.containers import Container
 from testcontainers.core.exceptions import ContainerIsNotRunning, NoSuchPortExposed
 from testcontainers.core.waiting_utils import WaitStrategy
 
@@ -137,9 +139,9 @@ class ComposeContainer:
         stdout, stderr = self._docker_compose.get_logs(self.Service)
         return stdout.encode(), stderr.encode()
 
-    def get_wrapped_container(self) -> "ComposeContainer":
+    def get_wrapped_container(self) -> Container:
         """Get the underlying container object for compatibility."""
-        return self
+        return self._docker_compose._get_docker_client().containers.get(self.ID)
 
     def reload(self) -> None:
         """Reload container information for compatibility with wait strategies."""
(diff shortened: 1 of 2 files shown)


alexanderankin:
ok, have to think about this one more carefully. seems like there is more to this than appears

core/testcontainers/compose/compose.py
```diff
@@ -137,9 +139,9 @@ def get_logs(self) -> tuple[bytes, bytes]:
         stdout, stderr = self._docker_compose.get_logs(self.Service)
         return stdout.encode(), stderr.encode()
 
-    def get_wrapped_container(self) -> "ComposeContainer":
+    def get_wrapped_container(self) -> Container:
         """Get the underlying container object for compatibility."""
-        return self
+        return self._docker_compose._get_docker_client().containers.get(self.ID)
```
surister: Dumb question: can `.containers.get(self.ID)` return a value other than `Container`, e.g. None? 

what if the container is not found?

alexanderankin:
it throws docker api exception or something. its in the docstr of the get
method. just navigate to it and you should see it also

PR #908, opened 2025-10-18
feat(core, postgres): Add RunFunctionWaitStrategy and use it in pg

jankatins:
I found no way to convert the current `container.exec()` based check function to any of the currently existing `WaitStartegies` and so I added `RunFunctionWaitStrategy` which can be used to convert such `container.exec()` based checks.

Helps with: #874
Closes: #883


jankatins:
I guess this can be closed after https://github.com/testcontainers/testcontainers-python/pull/935 was merged?

PR #915, opened 2025-11-04
fix(elasticsearch): Use wait strategy instead of deprecated decorator

artem-ilin:
Part of fixing https://github.com/testcontainers/testcontainers-python/issues/874.

Decided to remove `get_url` method as it's no longer used and is not part of the protocol.

### DIFF
--- a/core/testcontainers/core/wait_strategies.py
+++ b/core/testcontainers/core/wait_strategies.py
@@ -6,6 +6,7 @@ Structured wait strategies for containers.
 - HealthcheckWaitStrategy: Wait for Docker health checks to pass
 - PortWaitStrategy: Wait for TCP ports to be available
 - FileExistsWaitStrategy: Wait for files to exist on the filesystem
+- ExecWaitStrategy: Wait for command execution inside container to succeed
 - CompositeWaitStrategy: Combine multiple wait strategies
 
 Example:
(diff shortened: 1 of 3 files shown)


artem-ilin:
@alexanderankin whenever you have time please take a look

artem-ilin:
@alexanderankin kind reminder

artem-ilin:
@alexanderankin please take a look

alexanderankin (approved):

alexanderankin:
lets see the tests pass on python 3.9 for the module and then merge

artem-ilin:
@alexanderankin Huge thanks! Tests seem to pass. The codecov and docs/build jobs fail in the main branch.

maspetsberger:
Is there a specfic reason why the `get_url` function was removed?
It was a quite convenient way for downstream users to connect to, and pretty much all other containers offer a comparable feature.

alexanderankin:
i would probably say that this is because it is not specific to this subclass of Container, but that is not great, we can bring that back.

artem-ilin:
@maspetsberger I apologize for the inconvenience.
My motivation was that the method is not used internally anymore and was not documented.
Probably it was not a good idea to just remove it, without deprecation or anything.
Anyway, you can use `f"http://{es.get_container_host_ip()}:{es.get_exposed_port(es.port)}"` as a hot fix.

maspetsberger:
No worries Artem, I got it sorted on my end - after all it is not that complex of a function.

If you can put it in the backlog to bring it back, that would be appreciated. 
I think it would be good for convenience as well as architecture (since it allows for hiding the internals).

Thanks for responding so quickly.

PR #920, opened 2025-11-14
fix(core): restore dind test

alexanderankin:
I accidentally pushed this code in the previous commit, this PR associates itself with the change so the git history makes sense later. as well as fix the bug


alexanderankin:
@CarliJoy thoughts before i merge?

CarliJoy:
LGTM.

I am not sure that we require logger output within tests instead of simple print. But this is a question of taste.

alexanderankin:
closing in favor of #960

PR #922, opened 2025-11-14
fix: add PEP 561 py.typed markers for type checker support

andres-ito-traversal:
Add py.typed marker files to core and all module packages to indicate
  type information is available. This enables type checkers like Pyright
  and mypy to recognize and validate type hints in testcontainers packages.

  Resolves "Stub file not found" errors when running type checkers on
  code that imports testcontainers modules.

### DIFF
--- a/core/testcontainers/py.typed
+++ b/core/testcontainers/py.typed
@@ -1,0 +1,1 @@
+partial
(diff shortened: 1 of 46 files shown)


Tranquility2:
Already done as part of #1045

PR #918, opened 2025-11-12
fix: add PEP 561 py.typed markers for type checker support

andres-ito-traversal:
## Summary

  Adds `py.typed` marker files to all testcontainers packages to comply with [PEP 561](https://peps.python.org/pep-0561/) and enable type checking support.

  ## Motivation

  Without `py.typed` markers, type checkers like Pyright and mypy report "Stub file not found" errors when importing testcontainers modules, even though the code has inline type annotations.

  ## Changes

  - Added `py.typed` marker to `core/testcontainers/`
  - Added `py.typed` markers to all module packages under `modules/*/testcontainers/`
  - Updated `pyproject.toml` to ensure `py.typed` files are included in distributions

  ## Testing

  Verified that type checkers no longer report missing stub file errors:

  ```bash
  pyright -c "from testcontainers.postgres import PostgresContainer"
  # 0 errors, 0 warnings, 0 informations
```

##  Impact

  - Enables proper type checking for projects using testcontainers
  - No breaking changes or runtime behavior modifications
  - Aligns with Python packaging best practices (PEP 561)

Github Issue=https://github.com/testcontainers/testcontainers-python/issues/305

PR #927, opened 2025-11-21
fix(azurite): make visible to type checkers

### LINKED ISSUE #926: `py.typed` missing from the package
The [code for `azurite`](https://github.com/testcontainers/testcontainers-python/blob/main/modules/azurite/testcontainers/azurite/__init__.py) appears to be properly type-annotated, but the package does not include PEP-561 marker (`py.typed`). This prevents `mypy` from recognising `testcontainers` imports as being typed.

Expected: `py.typed` to be included in the package.

This is a narrow-scoped variant of #305

kveretennicov:
Closes #926 by adding PEP-561 marker file `py.typed`


edgarrmondragon:
This is a duplicate of https://github.com/testcontainers/testcontainers-python/pull/918

kveretennicov:
> This is a duplicate of #918

#918 has a global remit; this is concerned only with `azurite`, which, perhaps, will make it easier to land.

Tranquility2 (approved):

PR #941, opened 2025-12-22
fix(redis): Use wait strategy instead of deprecated decorator

rmnigm:
Part of fixing https://github.com/testcontainers/testcontainers-python/issues/874.


alexanderankin:
believe this has already been fixed

PR #944, opened 2026-01-01
feat(core): Pass-through extra execution arguments to the underlying container

libohad-dev:
Allow setting additional parameters on commands sent to the container, e.g. working directory or tailored environment variables.

### DIFF
--- a/core/testcontainers/core/container.py
+++ b/core/testcontainers/core/container.py
@@ -325,10 +325,10 @@ class DockerContainer:
             return "not_started"
         return cast("str", self._container.status)
 
-    def exec(self, command: Union[str, list[str]]) -> ExecResult:
+    def exec(self, command: Union[str, list[str]], *args: Any, **kwargs: Any) -> ExecResult:
         if not self._container:
             raise ContainerStartException("Container should be started before executing a command")
-        return self._container.exec_run(command)
+        return self._container.exec_run(command, *args, **kwargs)
 
     def wait(self) -> int:
         """Wait for the container to stop and return its exit code."""


alexanderankin:
there should be an api to let people do this but this is not it

libohad-dev:
Thanks for reviewing this. Understood that this isn't the right form for it.

I'll share what I was aiming at, so I can head toward the API you'd actually prefer. `exec()` here is a thin forwarder to `docker-py`'s `exec_run()`, and the options I was trying to reach — working directory, environment, user, I/O options, and so on — are the ones `exec_run()` itself already exposes. Mirroring the signature on the method that wraps it felt like the natural home for them.

Happy to defer to whatever interface fits the project best, though. Could you say what you'd consider the right API here? Once I have a clearer view of the target I'll gladly implement it properly.

alexanderankin:
yeah so in java you can change the raw object sent to the docker api after all pre-processing by testcontainers library. 

https://github.com/testcontainers/testcontainers-java/blob/905a445ad8b701e4c531aa652e42179072a88d01/core/src/main/java/org/testcontainers/containers/GenericContainer.java#L1393

https://github.com/docker-java/docker-java/blob/main/docker-java-api/src/main/java/com/github/dockerjava/api/command/CreateContainerCmd.java

part of the problem here is that the python docker client is not owned by testcontainers but by docker and the api is different ish to the java one (not so easy to deal with as pojo/dataclass/namedtuple/etc). i plan to address in v5 rewrite but dont have time. a pr that implements something like the above linked files would, however, probably get merged and released though

libohad-dev:
Thanks for the clarification.
I'll see if I can get something like that ready for review in the coming days.

libohad-dev:
Opened as #1050 
PTAL when you have the time. Let me know if it matches your intentions, or if you have any changes you'd like me to make

PR #946, opened 2026-01-04
fix(core): recreate poetry lockfile with latest versions of libraries

alexanderankin:
updated python version to 3.10 for simpler lockfile

recreated lockfile

updated ci for python 3.10 and no more sphinx (readthedocs is broken) since docs are mkdocs (which are not ready)


CarliJoy:
Thanks. This looks good.

Dropping 3.9 Support is good, it is EOL for 3 month:
https://endoflife.date/python

Please merge, in order for other MRs to be able  rebase and have a working Pipeline.

PR #949, opened 2026-01-06
feat(lowkeyvault): added Lowkey Vault module

### LINKED ISSUE #948: New Container: Lowkey Vault

I would like to implement a Lowkey Vault container (an Azure Key Vault emulator which already has .Net and Java Testcontainers modules).

Docker container: https://hub.docker.com/r/nagyesta/lowkey-vault
Docker configuration: https://github.com/nagyesta/lowkey-vault/tree/main/lowkey-vault-docker
Lowkey Vault can be very useful for projects which need direct integration with Azure Key Vault as it can enable local development and testing.

Although the generic container can start the container, the configuration is far form trivial due to the fact that the container uses 2 ports and has some pitfalls we can help avoid with a custom container.

**Other references:**

I can provide a PR for this if you are okay with that...

nagyesta:
- Add Lowkey Vault container implementation
- Cover the new container with tests
- Add basic example for secret use

Resolves #948

### DIFF
--- a/modules/lowkey-vault/testcontainers/lowkeyvault/__init__.py
+++ b/modules/lowkey-vault/testcontainers/lowkeyvault/__init__.py
@@ -1,0 +1,116 @@
+import os
+from enum import Enum
+from typing import Any, NamedTuple, Optional
+
+import requests
+
+from testcontainers.core.container import DockerContainer
+from testcontainers.core.wait_strategies import LogMessageWaitStrategy
+
+# This comment can be removed (Used for testing)
+
+
+class StaticToken(NamedTuple):
+    """Represents an OAuth access token."""
+
+    token: str
+    """The token string."""
+    expires_on: int
+    """The token's expiration time in Unix time."""
+
+
+class StaticTokenCredential:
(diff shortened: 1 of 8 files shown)


modules/lowkey-vault/testcontainers/lowkeyvault/__init__.py
```diff
@@ -0,0 +1,95 @@
...
+    def _configure(self) -> None:
+        return
+
+    def get_connection_url(self, network_type: NetworkType = NetworkType.LOCAL) -> str:
+        if network_type == NetworkType.LOCAL:
+            return f"https://{self.get_container_host_ip()}:{self.get_exposed_port(self.api_port)}"
+        else:
+            return f"https://{self.container_alias}:{self.api_port}"
+
+    def get_imds_endpoint(self, network_type: NetworkType = NetworkType.LOCAL) -> str:
```
nagyesta: We could also add a method to set the environment variables automatically for the managed identity simulation (although we cannot create the `/var/opt/azcmagent/tokens/assumed-identity.key` key file automatically).

PR #953, opened 2026-01-19
fix: Allow passing in a custom wait strategy string in MySQL, Cassandra, Kafka and Trino

SubhadityaMukherjee:
Hey,

Just a small change in adding the ability to pass in a custom wait strategy message for MySQL. The current implementation had '.*: ready for connections.*: ready for connections.*' as a default regex with no option to change it. This PR moves it into a parameter. 
No new functionality has been added so I presume the tests can stay the same.

If you find this useful, I can also go over the other implementations and make a similar change wherever this is a problem.

### DIFF
--- a/modules/cassandra/testcontainers/cassandra/__init__.py
+++ b/modules/cassandra/testcontainers/cassandra/__init__.py
@@ -39,7 +39,9 @@ class CassandraContainer(DockerContainer):
     CQL_PORT = 9042
     DEFAULT_LOCAL_DATACENTER = "datacenter1"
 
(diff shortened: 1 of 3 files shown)


SubhadityaMukherjee:
Okay I did a grep and added similar changes in Cassandra, Kafka and trino where it seemed to also need this as input. As before, I didnt change any functionality so I left the tests unchanged. 

Thanks :)

I needed to use testcontainers for a project and tried to debug this for a few hours before getting frustrated and just trying to patch the module instead.

alexanderankin (approved):

PR #973, opened 2026-03-02
Parametrize timeout in .wait_for() call

alexberndt:
Parametrize the timeout time in the `DockerCompose().wait_for()` method, sticking to the original default of `120` seconds.


alexanderankin:
ok so better if we can add this to the base strategy class, because you can have other parameters, like interval and frequency. if you must add this now, mark it as deprecated with a note that this will be a common field in all strategies in the future. like we can even keep this parameter around indefinitely because ergonomics is important but architecturally it is out of place.

Issue #964, opened 2026-02-04
Bug: Keycloak health check not working with HTTP relative path


KatharinaGoerlach:
**Describe the bug**

In case a HTTP relative path is configured the health check requires to consider this relative path. 
The Keycloak module does support port number and hard-coded health endpoint but not HTTP relative path during readiness probe (ie Keycloak health check).

(Related topic: With Keycloak version 26.3 the health endpoint is only provided via managament port but not other ports.)

**To Reproduce**

```python
from testcontainers.keycloak import KeycloakContainer

keycloak = KeycloakContainer('quay.io/keycloak/keycloak:latest') \
        .with_env("KEYCLOAK_ADMIN", 'test') \
        .with_env("KEYCLOAK_ADMIN_PASSWORD", 'test') \
        .with_env("KC_HTTP_RELATIVE_PATH", "/auth") \
        .with_env("KC_HEALTH_ENABLED", 'true') \
        .with_env("KC_HTTP_MANAGEMENT_RELATIVE_PATH", "/auth") \
        .with_command("start-dev")

keycloak.start()
```
		
		
With this Keycloak Container the readiness probe is failing with HTTP error

```python
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: http://localhost:53741/health/ready
```

However, keycloak health check is successful when HTTP relative path is included,i.e. when the following request is used:
```python
http://localhost:53741/auth/health/ready
```

For readiness probe in testcontainer please see 
https://github.com/testcontainers/testcontainers-python/blob/main/modules/keycloak/testcontainers/keycloak/__init__.py#L89

**Runtime environment**

Windows System
python 3.13.3
docker 28.4.0
testcontainers-python 4.14.0

themavik:
Fix submitted in #988.

PR #982, opened 2026-03-13
feat(keycloak): support for relative path and management relative path

JotBePunkt:
When these parameters were set, the functions get_url() and get_management_url() did return the wrong URLs (without the prefix)

Now the implementation checks if the parameter is checked and adds it to the returned URL

This solves https://github.com/testcontainers/testcontainers-python/issues/964


JotBePunkt:
@alexanderankin it would be great, if this change is in one of the next releases. What do we need to do to have this PR merged?

alexanderankin:
i started rebasing a bigger PR yesterday, working through merging all the things

Issue #983, opened 2026-03-16
Issue with testing ryuk under dind


Tranquility2:
We seem to have an issue that is not easy to reproduce with ryuk under dind.
the issue started to appear after I upgraded to Docker 29.3.0
```Server: Docker Engine - Community
 Engine:
  Version:          29.3.0
  API version:      1.54 (minimum version 1.40)
  Go version:       go1.25.7
  Git commit:       83bca51
  Built:            Thu Mar  5 14:25:48 2026
  OS/Arch:          linux/amd64
  Experimental:     false
```
Example from the tests:
```
=================================== FAILURES ===================================
_____________________________ test_wait_for_reaper _____________________________
core/tests/test_ryuk.py:46: in test_wait_for_reaper
    with pytest.raises(NotFound):
E   Failed: DID NOT RAISE <class 'docker.errors.NotFound'>
------------------------------ Captured log call -------------------------------
INFO     testcontainers.core.container:container.py:178 Pulling image testcontainers/ryuk:0.8.1
INFO     testcontainers.core.container:container.py:207 Container started: 11c29242a3a6
INFO     testcontainers.core.container:container.py:178 Pulling image hello-world
INFO     testcontainers.core.container:container.py:207 Container started: 0faee54fa35a
_________________________ test_container_without_ryuk __________________________
.venv/lib/python3.10/site-packages/docker/api/client.py:275: in _raise_for_status
    response.raise_for_status()
.venv/lib/python3.10/site-packages/requests/models.py:1026: in raise_for_status
    raise HTTPError(http_error_msg, response=self)
E   requests.exceptions.HTTPError: 409 Client Error: Conflict for url: http+docker://localhost/v1.54/containers/11c29242a3a66f2fc35fe87d8ccdf169c05af8b3a946bb823b7aa295b9bde314?v=True&link=False&force=True

The above exception was the direct cause of the following exception:
core/tests/test_ryuk.py:58: in test_container_without_ryuk
    Reaper.delete_instance()
core/testcontainers/core/container.py:333: in delete_instance
    Reaper._container.stop()
core/testcontainers/core/container.py:212: in stop
    self._container.remove(force=force, v=delete_volume)
.venv/lib/python3.10/site-packages/docker/models/containers.py:367: in remove
    return self.client.api.remove_container(self.id, **kwargs)
.venv/lib/python3.10/site-packages/docker/utils/decorators.py:19: in wrapped
    return f(self, resource_id, *args, **kwargs)
[...]

PR #984, opened 2026-03-16
fix(core): wait for ryuk more reliably, improve tests: long_running, filter logs

Tranquility2:
This was originally only to fix https://github.com/testcontainers/testcontainers-python/issues/983 but I took the time to fix some other stuff that made reading and running the tests less optimal.

- Fix test_ryuk flakiness: replace fixed sleep with polling helper (_wait_for_container_removed) that waits for Ryuk to finish reaping
- Suppress expected error logs in test_compose_volumes and test_wait_strategies to reduce test noise
- Add __test__ = False to TestcontainersConfiguration to prevent pytest from trying to collect it as a test class
- Add `long_running` pytest marker to pyproject.toml
- Mark DinD/DooD tests as long_running since they build Docker images and take x4 than all the other tests combined 
- Add `quick-core-tests` Makefile target to run core tests excluding long_running tests for faster feedback loops


Tranquility2:
Looks like there is an issue with `cassandra-driver` (v3.29.3) in python 3.14
This blocks:
1. linting
2. test with python 3.14

Tranquility2:
@alexanderankin mind taking a look?

PR #993, opened 2026-03-29
feat(core): support SSH-based DOCKER_HOST

### LINKED ISSUE #992: SSH-based DOCKER_HOST not supported
## Description

When using an SSH-based `DOCKER_HOST` (e.g. `ssh://user@remote-host`), testcontainers cannot connect to containers.

The Docker Python SDK rewrites SSH URLs internally to `http+docker://ssh`, losing the original hostname. `DockerClient.host()` then parses that rewritten URL and returns `"ssh"` instead of the actual remote hostname. This breaks `get_container_host_ip()`, `get_exposed_port()`, and any code that relies on the host address to reach containers.

## Reproduction

```bash
export DOCKER_HOST="ssh://user@10.0.0.1"
python -c "from testcontainers.core.docker_client import DockerClient
print(DockerClient().host())"
```

Expected: 10.0.0.1

Actual: ssh

Tranquility2:
Fixes #992 

## Problem

When `DOCKER_HOST` is set to an SSH URL (e.g. `ssh://user@remote-host`), the Docker Python SDK rewrites `base_url` to `http+docker://ssh`, losing the original hostname. This causes `DockerClient.host()` to return `"ssh"` instead of the actual remote address, breaking container connectivity.

Additionally, the SDK defaults to paramiko for SSH connections, which crashes under pytest due to stdin capture conflicts.

### `docker_client.py`
- Extract the remote hostname from `DOCKER_HOST` in `host()` instead of relying on the SDK's rewritten `base_url`
- Default to `use_ssh_client=True` for SSH connections to avoid paramiko/pytest stdin conflicts
- Sanitize SSH URLs with unsupported path components before passing to the SDK
- Add `get_docker_host_hostname()` and `is_ssh_docker_host()` helpers

### `compose.py`
- Handle SSH in `PublishedPortModel.normalize()` — replace local bind addresses (`0.0.0.0`, `127.0.0.1`, etc.) with the remote SSH hostname

### Tests
- Add SSH-specific tests for `DockerClient.host()`, connection mode, compose port normalization, and URL sanitization


Tranquility2:
image

PR #1007, opened 2026-04-15
fix(core): add container logs to TimeoutError when wait strategy times out

ticapix:
Hello,

Thanks for testcontainers !

I had a lot of trouble to understand why my docker strategy was timing out and I ended up patching the code like in this PR to get the container logs when there is a timeout.

I hope it can help others when debugging.

`make tests` runs ok

Happy to update the PR if needed


alexanderankin (approved):

alexanderankin:
not sure why the cicd is not running 👀

alexanderankin:
core/testcontainers/core/container.py:223: error: "TimeoutError" has no
attribute "add_note"  [attr-defined]
                    ex.add_note(self._container.logs().decode())
                    ^~~~~~~~~~~
Found 1 error in 1 file (checked 45 source files)

PR #1018, opened 2026-05-29
fix(cosmosdb): replace distutils.strtobool removed in Python 3.12

FidelusAleksander:
## Problem

The `cosmosdb` module imports `distutils.util.strtobool` in `_emulator.py`. The `distutils` module was deprecated in Python 3.10 ([PEP 632](https://peps.python.org/pep-0632/)) and **removed in Python 3.12**. This causes an `ImportError` when using the CosmosDB module on Python 3.12+:

```
ModuleNotFoundError: No module named 'distutils'
```


Tranquility2:
LGTM

Tranquility2 (approved):

PR #1029, opened 2026-06-02
test(core): decouple registry tests from testcontainers.registry

Tranquility2:
Following @alexanderankin's feedback on the previous approach.

`core/tests/test_core_registry.py` was importing `DockerRegistryContainer` from a sibling module.
That couples `core` to a sibling, which becomes a hard wall once the planned `core` / `community` split lands.

This PR keeps the integration coverage but inlines a private `_LocalRegistryContainer` helper into `core/tests/`.
The duplication is intentional and private to `core/tests/`.

Skip markers (macOS, Podman, SSH-based Docker) stay as they were on `main`.
No production code changes.

### DIFF
--- a/core/tests/test_core_registry.py
+++ b/core/tests/test_core_registry.py
@@ -1,29 +1,25 @@
-"""Integration test using login to a private registry.
+"""
+Integration tests for private-registry support in ``DockerClient``.
 
-Note: Using the testcontainers-python library to test the Docker registry.
-This could be considered a bad practice as it is not recommended to use the same library to test itself.
-However, it is a very good use case for DockerRegistryContainer and allows us to test it thoroughly.
+These tests spin up a real ``registry:2`` container and exercise the wiring that turns ``DOCKER_AUTH_CONFIG`` into a successful login + image pull.
+The registry is provided by the local ``_LocalRegistryContainer`` helper, kept as a private copy in ``core/tests`` so ``core`` does not import from a sibling module.
 
-Note2: These tests are skipped on macOS and SSH-based Docker hosts because they rely on insecure HTTP registries,
-which are not supported in those environments without additional configuration.
+Skipped where insecure HTTP registries cannot be reached without daemon reconfiguration.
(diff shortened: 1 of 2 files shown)


core/tests/test_core_registry.py
```diff
@@ -1,103 +1,107 @@
...
-However, it is a very good use case for DockerRegistryContainer and allows us to test it thoroughly.
+These tests previously relied on a real ``DockerRegistryContainer`` (an
+integration test that pushed/pulled images against a live ``registry:2``).
+They now use a mocked Docker SDK (see ``registry_mocks.py``) so they run
+offline and cover only the logic that ``testcontainers`` itself owns:
 
-Note2: These tests are skipped on macOS and SSH-based Docker hosts because they rely on insecure HTTP registries,
-which are not supported in those environments without additional configuration.
+* logging in to a registry using credentials decoded from ``docker_auth_config``, and
+* surfacing ``docker.errors.NotFound`` when an image is missing on pull.
```
alexanderankin: i think this is not great actually, integration testing is good, hmmmm.

alexanderankin: but then again, maybe this was too clever.

alexanderankin: i think i would rather solve this by copy and pasting the code between them. @CarliJoy has been bugging me to do a `src/` directory reorganization, which is now also a split into a core and community package (cannot split into individual packages directly at the moment because of a previous failed attempt to do so and now we do not control the packages on pypi). but then we can evolve core and now its less of a big scary monolith. so probably lets just copy and paste the relevant parts of the DockerRegistryContainer (really the distribution registry) class and maybe call it the TestDockerRegistryContainer instead or whatever.

alexanderankin (changes requested):

core/tests/test_core_registry.py
Tranquility2: Will do

alexanderankin (approved):

PR #1031, opened 2026-06-03
chore(deps): Update locking and enable auto-updating

CarliJoy:
Updated the lockfile once manually with uv lock --upgrade. Also added a dependabot configuration that should performe these updates in the future.

See https://docs.astral.sh/uv/guides/integration/dependabot/


alexanderankin (approved):

Tranquility2:
Super nice

PR #978, opened 2026-03-06
feat(mqtt): MosquittoContainer: Add version 2.1.2

f18m:
This PR enables the testcontainer MosquittoContainer (which I contributed to this project a few years ago) to work with version 2.1.2 and higher of Mosquitto.

In particular this PR is fixing an issue where rw mode is now needed for /data partition for Mosquitto version 2.1.1 (released 2026-02-04), which fixed a PUID/PGID issue.
In addition, it contains a fix for Mosquitto warnings generated by the `listener` directive appearing after `protocol` directive.

PR #1034, opened 2026-06-04
fix(k3s): use with_tmpfs_mount to avoid duplicate tmpfs kwarg

Tranquility2:
## Problem

`K3SContainer.__init__` configures tmpfs mounts via `with_kwargs`:

```python
self.with_kwargs(privileged=True, tmpfs={"/run": "", "/var/run": ""})
```

Since #978, `DockerContainer.start()` also passes `tmpfs=self.tmpfs` explicitly to `docker_client.create()`, while still spreading `**self._kwargs`. This causes a duplicate-keyword error on every K3S container start:

```
TypeError: docker.models.containers.ContainerCollection.run() got multiple values for keyword argument 'tmpfs'
```

CI evidence: every `test (3.10|3.13|3.14, k3s)` job fails with this error.

## Fix

Migrate to the dedicated `with_tmpfs_mount()` API introduced in #978, leaving only `privileged=True` in `with_kwargs`.

## Verification

`uv run pytest -q modules/k3s/tests/test_k3s.py` no longer raises the `TypeError` (container starts successfully; remaining test progression depends on CI environment as before).

### DIFF
--- a/modules/k3s/testcontainers/k3s/__init__.py
+++ b/modules/k3s/testcontainers/k3s/__init__.py
@@ -44,7 +44,9 @@ class K3SContainer(DockerContainer):
         self.with_exposed_ports(self.KUBE_SECURE_PORT, self.RANCHER_WEBHOOK_PORT)
         self.with_env("K3S_URL", f"https://{self.get_container_host_ip()}:{self.KUBE_SECURE_PORT}")
         self.with_command("server --disable traefik --tls-san=" + self.get_container_host_ip())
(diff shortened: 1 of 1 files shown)


Tranquility2:
Superseded by 4d1da0cc on main (`fix(k3s): replace tmpfs dict kwarg with with_tmpfs_mount()`).
`src/testcontainers/community/k3s/__init__.py` now uses `with_tmpfs_mount("/run")` and `with_tmpfs_mount("/var/run")`, which addresses the same duplicate `tmpfs` kwarg issue.
Closing.

PR #1035, opened 2026-06-04
test(milvus): cast exposed_port to str for substring assert

Tranquility2:
## Problem

`test_run_milvus_success` asserts:

```python
assert url and exposed_port in url
```

`get_exposed_port()` returns `int`, so the `in` check against a string `url` raises:

```
TypeError: 'in <string>' requires string as left operand, not int
```

CI evidence: `test (3.10|3.13|3.14, milvus)` jobs fail this assertion for both `v2.4.0` and `v2.4.4`.

## Fix

Cast `exposed_port` to `str` before the substring check.


Tranquility2:
Superseded by the src-layout migration (#1045) and subsequent test fixes on main.
`tests/community/milvus/test_milvus.py` already does `assert str(exposed_port) in url`, which is the same fix proposed here.
Closing.

PR #1036, opened 2026-06-04
chore(ci): skip scylla module tests on Python 3.14

Tranquility2:
## Problem

`test (3.14, scylla)` fails with:

```
ModuleNotFoundError: No module named 'cassandra'
```

`cassandra-driver` is gated by `python_version < '3.14'` in two places in `pyproject.toml`:

```toml
scylla = ["cassandra-driver>=3; python_version < '3.14'"]
...
test = ["cassandra-driver>=3; python_version < '3.14'", ...]
```

So on Python 3.14 the driver isn't installed, but the matrix still attempts to run scylla tests, which import `cassandra.cluster` at runtime inside `ScyllaContainer._connect()`.

## Fix

Exclude the `(3.14, scylla)` combination from the test matrix to reflect the existing dependency constraint.

## Note

The `cassandra` module has the same dependency constraint and may need the same treatment, but is left out of this PR since CI does not currently flag it (likely not in `changed_modules` for recent PRs).

### DIFF
--- a/.github/workflows/ci-community.yml
+++ b/.github/workflows/ci-community.yml
@@ -50,6 +50,11 @@ jobs:
       matrix:
         python-version: ["3.10", "3.13", "3.14"]
         module: ${{ fromJSON(needs.track-modules.outputs.changed_modules) }}
+        exclude:
+          # cassandra-driver has no Python 3.14 support
+          # (see `python_version < '3.14'` markers in pyproject.toml).
+          - python-version: "3.14"
+            module: scylla
     steps:
       - name: Checkout contents
         uses: actions/checkout@v4


Tranquility2:
Closing as obsolete.

The community CI workflow was restructured in #1045 to use `dorny/paths-filter` with a per-module matrix, so the `exclude` strategy proposed here no longer applies cleanly.

The underlying ScyllaContainer / cassandra-driver incompatibility on Python 3.14 is now partially handled by a marker on the optional dep — `scylla = ["cassandra-driver>=3; python_version < '3.14'"]` — but the tests themselves can still fail on 3.14 if exercised.
If that turns out to be a recurring CI problem, it would need a fresh approach against the new matrix layout (e.g., skipping the scylla module entry for 3.14 in `compute_modules.py` or marking the tests with `@pytest.mark.skipif`).

PR #1038, opened 2026-06-05
feat(lint): replace mypy with ty type checker

### LINKED ISSUE #1001: Utilizing Ty astype checker (replacing mypy)
> Maybe we can first move from mypy to [Ty ](https://docs.astral.sh/ty/)?
> *since we are already on the Astral bandwagon 😅

as mentioned on #959

Tranquility2:
Closes #1001

Swaps mypy for [Astral's ty](https://github.com/astral-sh/ty) as the project's static type checker.
ty is ~20x faster on this codebase (0.25s vs 4.9s locally) and provides clearer diagnostics.

## Tooling changes

- `pyproject.toml`: drop `mypy>=1` and `types-paramiko>=4` from the lint dep group, add `ty`.
  Replace the `[tool.mypy]` block + `[[tool.mypy.overrides]]` entries with `[tool.ty.*]` blocks.
  `unresolved-import` and `possibly-missing-submodule` are set to `ignore` to preserve the previous `ignore_missing_imports` behavior for `docker` / `wrapt` / `requests` / `registry`.
- `Makefile`: replace `mypy-core` / `mypy-core-report` with `ty-core`.
- `.pre-commit-config.yaml`: swap the local mypy hook for a local ty hook.
- `.github/workflows/ci-lint.yml`: run `make ty-core`.
- `scripts/mypy_report.py`: removed (no longer needed; ty's own output is concise).
- `README.md`: add ty badge.

## Code fixes for diagnostics surfaced by ty

- `container._transfer_into_container`: default `mode=0o644` to match `with_copy_into_container`.
- `container.get_wrapped_container` / `image.get_wrapped_image`: annotate return as `Optional[...]` (they can be `None` before `start()`).
- `wait_strategies._get_status_tc_container`: raise `ContainerStartException` when the container is not started instead of dereferencing `None`.
- `waiting_utils`: use `getattr(self.func, "__name__", repr(self.func))` for safer error messages.
- Targeted `# ty: ignore[...]` for the hasattr-guarded `add_note`, the Protocol-narrowed `container.exec` call, the kwargs unpack into docker-py, and the wrapt decorator cast.

## Verification

- `uv run ty check core/testcontainers` → **All checks passed!**
- `make ty-core` → passes
- `uv run pytest -q core/tests` → 121 passed


Tranquility2:
Closing this in light of the src-layout migration (#1045) which expanded mypy coverage to the whole package and explicitly noted mypy now passes for everything.

After rebasing onto main and running ty against the new `src/testcontainers/` scope, ty reports **371 diagnostics** where mypy currently passes (~92 even scoped to just `src/testcontainers/core`).
The vast majority (342) are `invalid-argument-type` errors driven by `**kwargs`/`**dict` unpacking patterns that mypy permissively accepts but ty strictly type-checks.
Dropping `types-docker` barely helps (371 → 362), so it's not a stubs issue — it's an inherent strictness gap.

Making ty green would require either hundreds of per-call `# ty: ignore` suppressions (lots of churn, low signal) or globally relaxing `invalid-argument-type` (which removes much of ty's value over mypy).
Combined with ty still being beta (v0.0.44), the swap isn't a clean drop-in today.

Worth revisiting once ty either:
- matures past beta and improves `**kwargs` inference, or
- the codebase migrates to fully-typed kwargs (TypedDict / explicit params).

Closes without merging.
Issue #1001 can stay open as a tracking item, or be closed.

PR #1045, opened 2026-06-05
feat(main): Migrate to src layout

### LINKED ISSUE #305: Mark packages as typed

### LINKED ISSUE #720: Feature: Change to another layout [src]
Follow up on the discussion in #488.

The MR suggested to change to a flat layout to activate typing easily (see #305).

@totallyzen  brought the [argument](https://github.com/testcontainers/testcontainers-python/pull/488):

> Since I wrote the refactor to the current core and modules, I can chime in a bit.
> 
> There is a very simple reason why it is core and modules -> it is because all major testcontainers language flavors use these. I don't think I'd want to deviate from that as it allows the TC mainline documentation to converge and easily match support across languages (automation can be built for it, and issues tracked across language flavours)
>
> I wish we had better automation and support from the TC staff (I'm only a volunteer myself like @alexanderankin) but this is what we've got for now.

However, after looking at the official [.NET project](https://github.com/testcontainers/testcontainers-dotnet), I find this argument weak. The .NET project doesn't follow this structure—there is no core or modules, only src and tests folders.

The current layout introduces several issues:

1. It is complex and hard to understand, as only with some Poetry-specific setup are the different folders combined.
2.   It makes running poetry pytest to test the project nearly impossible due to "import file mismatches."
3.  It suggests the existence of subprojects like [testcontainers-core](https://pypi.org/project/testcontainers-core/) (which actually exists on PyPi, created by @tillahoffmann), adding confusion.
4.    Properly type-checking the project is difficult because mypy expects the entire project (not only parts) to be typed when importing, and it checks for the existence of `py.typed`.

I propose adopting the [src-layout](https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/), which has been embraced by many modern Python packages. This would:

   -  Clarify the difference between tests for the final package and tests in "development mode."
   - Allow running pytest from the project root, reducing the need for a complex Makefile just to run tests. Tests for different modules can still be organized into subfolders in the tests directory.

Most importantly, this change would reduce complexity, enabling Python developers to contribute more easily with less cognitive overhead.

Lastly, it would mirror the structure of the official .NET project. :)

### LINKED ISSUE #968: request: change repo layout (to src based layout)
any issues or pull requests opened after this one are going to need to adapt to the new layout.

for any issues or pull requests already open: i will either merge or adapt them myself.

CarliJoy:
This merge requests changes the layout of testcontainers to use `src` layout. 

Doing this some bugs surfaced that were fixed.
Also some minor improvement were made on the way. Hopefully everything is still readable.
I also used the opportunity to clean up the ruff config a bit and activate it again for tests.

Also the complete package is now typed and mypy passes 🎉 

Closes #720 
Closes #968
Closes #305 

Please don't squash this mr or the commit history will be hard to understand.


pyproject.toml
```diff
@@ -375,15 +301,15 @@ check_untyped_defs = true
 disable_error_code = ["no-untyped-def"]
 
 [[tool.mypy.overrides]]
-module = ["docker.*"]
+module = ["pika.*"]
 ignore_missing_imports = true
```
CarliJoy, line 379: Fun fact: Docker was (probably) never really used to type check against, because its stubs were missing. 

At the same time it is the most important library for this project ;-)

alexanderankin (approved):
lets get this over with 👍

Tranquility2:
Amazing work @CarliJoy

PR #1050, opened 2026-06-10
feat: extended configuration options for container.exec method

libohad-dev:
The branch adds the `ExecConfig` struct (patterned on the [Java implementation of the same feature](https://github.com/testcontainers/testcontainers-java/blob/main/core/src/main/java/org/testcontainers/containers/ExecConfig.java)) and includes it as a valid input for `DockerContainer.exec()`. This allows setting the running user, working directory, environment variables, and privileged mode. Other features become easy extensions of the interface.
Also included in the branch:
- Extensions to the core test suite, pinning existing behavior of `.exec()` that wasn't previously tested, notably the fact that `str`-typed commands _are not_ executed in shell mode
- Rewrite of `docs/features/executing_commands.md`, which incorrect prior to this branch. It now aligns with the actual code

### DIFF
--- a/docs/features/executing_commands.md
+++ b/docs/features/executing_commands.md
@@ -4,104 +4,80 @@ Testcontainers-Python provides several ways to execute commands inside container
 
 ## Basic Command Execution
 
-The simplest way to execute a command is using the `exec` method:
+The simplest way to execute a command is using the `exec` method, passing either an argv list or a string:
 
 ```python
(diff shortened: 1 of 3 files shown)


alexanderankin (approved):

alexanderankin:
didnt realize it was just for the exec method. that is pretty minor so i think its fine

